PR #57, opened 2015-01-29
Expose mangle config option

benib:
This allows to set `minify: true, mangle: false` in config. `mangle` defaults to `true` to not break bc.
This is needed if you use angular and want to use Implicit Annotation e.g.

Being able to remove whitespace and comments during the build is probably still desirable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,12 @@ As well as an `options.config` parameter, it is also possible to specify minific
   builder.build('myModule', 'outfile.js', { minify: true, sourceMaps: true, config: cfg });
 ```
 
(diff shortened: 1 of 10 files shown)


guybedford:
This looks great! Any chance you can rebase onto the latest changes?

guybedford:
Btw, was something specific not working that you needed to add the mangle option?

SekibOmazic:
+1 for this PR!

benib:
Ok. I kind of messed up with the rebasing. I will send in another PR.

benib:
@guybedford I wanted to use angular's implicit annotation. As the mangling messes this up, I needed the option.

PR #59, opened 2015-01-29
expose mangle option, allows for minifying without mangling

benib:
This PR is the same as #57 without the rebase mess.

This allows to set minify: true, mangle: false in config. mangle defaults to true to not break bc.
This is needed if you use angular and want to use Implicit Annotation e.g.

Being able to remove whitespace and comments during the build is probably still desirable.


guybedford:
Thanks!

Issue #58, opened 2015-01-29
Source map paths are relative to `baseURL`, instead of `outPath`


guybedford:
The source map chaining in https://github.com/systemjs/builder/blob/master/compilers/es6.js#L59 seems to work correctly.

But I'm not sure why it isn't failing because the `sources` array contains module names like `file` instead of `file.js`.

Given this, how is Traceur knowing that the original file is `file.js`? Is it from `load.address`? That seems like the wrong assumption to be making?

crisptrutski:
Haven't look at that at all. Is this really a chaining issue though, perhaps I'm misunderstanding but sounds like the issue is with the vanilla traceur sourcemap or the input map (chaining should not generate any source names of its own)

guybedford:
@crisptrutski did you manage to look into this?

crisptrutski:
Took a quick look - those names are coming from `load.name` (https://github.com/systemjs/builder/blob/master/compilers/es6.js#L44). We don't seem to have the base path in scope here (we could work backwards from the name.. hacky). 

Seems we might as well just yank the extension (if any) from the address and just plonk that on. :goat:

guybedford:
Thanks, I'm wondering if I had my test case wrong - it should pick up the original source names from the first source file if chaining source maps being the point.

Would be nice to verify these are working correctly with some more test cases. Including if the original file was compiled from a few separate files into ES6, and then compiled into ES5, those first files should display as individual entries when debugging.

crisptrutski:
Sure - will write specs for `[6to5, traceur]` x `[single, multiple]` source map output cases, given fixture files.

guybedford:
Thanks that would be great.

crisptrutski:
Hijacking this issue for key issue. Will keep "relative to baseURL" as behaviour for in-memory builds

guybedford:
Perfect.

PR #63, opened 2015-02-09
Output to memory

### LINKED ISSUE #54: Output to memory
I don't see a howto, but hopefully there is a way to output to a memory object rather than a file in the FS. Or perhaps an option to set an output filesystem that could be a memory FS.

crisptrutski:
@guybedford let me know if you're happy with this approach, most notably the API (return file-contents string within promise). 

Perhaps a stream is more suitable, and I'm not sure what was meant by "memory FS". Looked at things like https://github.com/tschaub/mock-fs, but that sort of evil seems doesn't seem to belong outside of testing, and not sure the value it would add in the scenario vs return an inlined-file or a dictionary of the separate files.

Need to hook up Uglify for in-memory usage, and refactor my goop (been a while, JavaScript)

Closes #54

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -2,7 +2,8 @@ var saucy = require('./sourcemaps');
 var path = require('path');
 var mkdirp = require('mkdirp');
 var fs = require('fs');
-var asp = require('rsvp').denodeify;
+var rsvp = require('rsvp');
+var asp = rsvp.denodeify;
 var uglify = require('uglify-js');
 
 function countLines(str) {
@@ -36,7 +37,7 @@ function processOutputs(outputs) {
       if (map) {
         sourceMapsWithOffsets.push([offset + offset_, map]);
       }
-    } 
+    }
     else if (typeof output == 'string') {
       source = output;
     }
@@ -59,7 +60,7 @@ function createOutput(outputs, outFile, baseURL, createSourceMaps) {
 
   if (createSourceMaps && outputObj.sourceMapsWithOffsets.length)
     sourceMap = saucy.concatenateSourceMaps(outFile, outputObj.sourceMapsWithOffsets, baseURL);
-  
+
   var output = outputObj.sources.join('\n');
 
   return {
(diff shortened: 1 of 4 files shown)


lib/builder.js
```diff
@@ -120,12 +110,45 @@ exports.writeOutputFile = function(opts, outputs, baseURL) {
...
   });
+}
+
+function writeOutputString(opts, output, baseURL) {
+  var content = output.source;
+  if (opts.sourceMaps) {
+    content += '\n//# sourceMappingURL=data:application/json;base64,' +
+      new Buffer(output.sourceMap.toString()).toString('base64');
+  }
+  return wrapPromise(content);
```
guybedford: You can just do `Promise.resolve(content)` here.

lib/builder.js
```diff
@@ -120,12 +110,45 @@ exports.writeOutputFile = function(opts, outputs, baseURL) {
...
+  return new rsvp.Promise(function(resolve, reject) {
+    return resolve(value);
   });
+}
+
+function writeOutputString(opts, output, baseURL) {
+  var content = output.source;
+  if (opts.sourceMaps) {
+    content += '\n//# sourceMappingURL=data:application/json;base64,' +
+      new Buffer(output.sourceMap.toString()).toString('base64');
```
guybedford: I think even with a memory option output, it's fine to return the source map as a separate return value rather than encoding, as that's easier for chaining in custom builds.

guybedford:
:+1: perhaps just `output.source` and `output.sourceMap` though?

crisptrutski:
Yeah, makes the tests for #58 easier too.. 

Was originally intending to send to standard output for CLI usage, hence the inline-ing, but would need to disable all the other logging though, so perhaps reserve a flag for that.

Not sure if people would actually want to use it this way though.

guybedford:
Ok great. Yeah CLI streaming can be implemented if and when there is demand, happy to leave it out for now.

crisptrutski:
Points addressed, let me know if I should squash

guybedford:
Awesome, thanks. Are the source maps looking like they're working?

crisptrutski:
Looking good for non-chained. Will squash get this merged, look at the chaining in #64

PR #64, opened 2015-02-09
WIP Automated tests

crisptrutski:
(depends on in-memory stuff, will rebase once that's merged)


guybedford:
Woo!

crisptrutski:
@guybedford 

1. Not sure you'd like tests organized. Went with mocha since that's used in jspm-cli.
    Tests currently run via `mocha test/unit-tests.js` which is a bit clumsy, but otherwise runner will try choke on the other files in `test/`

2. The multiple-files + traceur output is generating some screwy paths. This regression happened during the writing of the tests (took a JSON dump right at the start). Any ideas why spamming `../`?

3. Will add tests exercising the chaining now, just made sense to start with the source files already there

crisptrutski:
Closes #58

guybedford:
Can you add this into the `tests` part of the package.json for an easy run with `npm run test:unit` for example?

guybedford:
Probably advisable to split up the tests into subdirectories now yes.

guybedford:
Are you running Traceur 0.0.82 - I believe there are source maps changes between 0.0.82 and 0.0.84 so we probably need to watch out for that.

crisptrutski:
Yeah I'm using 0.0.82. Do we need to support both or can we bump?

guybedford:
I'm happy to bump up and do a full release through all the projects, just watch the discussions on Traceur as this may change again.

PR #65, opened 2015-02-09
Updated README.md to show that it's possible to use minify, sourceMaps e...

markstickley:
...tc with `buildTree` as well as `build`


guybedford:
Thanks! Note that `opts.config` is only supported for `build` and `buildSFX`, not `buildTree`. Correcting that now.

PR #69, opened 2015-02-11
Builder class

crisptrutski:
Closes #44

PR #70, opened 2015-02-11
Fix 6to5 sourcemaps

### LINKED ISSUE #62: 6to5 source maps stopped working
It seems the sources paths aren't being set correctly anymore unfortunately.

crisptrutski:
Fixes #62 

@guybedford basically the issue is that that traceur emits a complete `file://` URL as the `sourceRoot`, whereas 6to5 is leaving that nil. 

When we concatenate the source maps, the `sourceRoot` gets baked in, and then our workaround for a Traceur quirk here was messing things up.

Another fix would be to fill in the `sourceRoot` earlier, but see no reason not to support relative URLs in general.


crisptrutski:
(includes #69)

crisptrutski:
@guybedford convenience link just to relevant commit: https://github.com/systemjs/builder/commit/ed653388a09173297183e836f5c7e1caf42ca448

PR #71, opened 2015-02-12
Remove handcoded compiler lookups and enumeration.

crisptrutski:
@guybedford behaviour-preserving refactor to remove confusion about which compilers require sfx wrapping, and to make the automatic runtime injection easier


guybedford:
:+1:

PR #73, opened 2015-02-12
Ensure `sources` paths relative to source map file

### LINKED ISSUE #58: Source map paths are relative to `baseURL`, instead of `outPath`

crisptrutski:
Closes #58

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -58,8 +58,10 @@ function createOutput(outputs, outFile, baseURL, createSourceMaps) {
 
   var outputObj = processOutputs(outputs);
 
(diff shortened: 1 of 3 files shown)


crisptrutski:
@guybedford let me know when I should back off on the self merges

guybedford:
All good.

Issue #77, opened 2015-02-14
buildSFX breaking in 8.1


silenceisgolden:
~~Pardon me if this is a stupid question, but I'm not exactly sure that buildSFX is actually calling 6to5. I've narrowed down my issue in my repo [6to5 errors](https://github.com/silenceisgolden/6to5-errors). What I am noticing is that `let` is IS NOT transformed into `var` when put through the build. When I use the 6to5 CLI on the `src/app.js` file in the repo, it shows the `let` IS being transpiled to `var`. This might just be me configuring something wrong, but there doesn't seem to be an example to do this a proper way with 6to5 and builderSFX.~~

Updated: Changed syntax of sample app code to be module-based instead of lazily treating it like 6to5 file compilation. This results in a different error `Uncaught: Module _____ not present`. I am reverting to 8.0 for now since the error is only in 8.1. I'll work on a pull request if I can learn the source enough to feel confident to make a change.

Thanks for adding support for 6to5, sorry to ping an issue so fast after the new releases.

silenceisgolden:
Traced this to https://github.com/6to5/6to5/issues/732, PR has been made and squashed https://github.com/systemjs/builder/pull/78.

crisptrutski:
@silenceisgolden thanks for all the investigative work you've done here - and sorry for breaking for your build! Getting over confident now that we have *some* test coverage.

Ideally the solution would "just work" through `sourceRoot` or `moduleRoot`, but in the mean time I'll just add back `filenameRelative`, just with less fudging (or, to be lazy for now, just some de-fudging)

crisptrutski:
@silenceisgolden mind trying with latest master? (see https://github.com/systemjs/builder/commit/6494379721396487ce1e736dc061aaed928c6d3c)

guybedford:
Thanks for reporting, fixed in 0.8.2.

silenceisgolden:
Thanks for fixing @crisptrutski and @guybedford. Much appreciated.

PR #78, opened 2015-02-14
reverting deletion of filenameRelative from es6 compiler

silenceisgolden:
Fixes https://github.com/systemjs/builder/issues/77, see commit comment for details.


silenceisgolden:
Turns out this doesn't fix the issue, so I'm closing this. See #77 for details.

PR #81, opened 2015-02-18
Arithmetic composition comes into builder

### LINKED ISSUE #80: Bundle arithmetic
jspm provides a bundle arithmetic syntax handled in https://github.com/jspm/jspm-cli/blob/master/lib/bundle.js#L113.

It could be worth implementing this arithmetic as a class method on the builder. `builder.bundleExpression` perhaps?

crisptrutski:
Closes #80

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,8 +79,8 @@ Builder.prototype.lookupOperatorFn = function(symbol) {
     throw 'Unknown operator ' + op.operator;
 };
 
-Builder.buildExpression = function(firstModule, operations, cfg) {
-  builder = new Builder();
+Builder.prototype.buildExpression = function(firstModule, operations, cfg) {
+  var builder = this;
 
   return Promise.resolve(builder.trace(firstModule, cfg))
   .then(function(trace) {
--- a/index.js
+++ b/index.js
@@ -70,6 +70,42 @@ Builder.prototype.build = function(moduleName, outFile, opts) {
   });
 };
 
+Builder.buildExpression = function(firstModule, operations, cfg) {
+  builder = new Builder();
+
+  return Promise.resolve(builder.trace(firstModule, cfg))
+  .then(function(trace) {
+    // if there are no other operations, then we have the final tree
+    if (!operations.length)
+      return trace.tree;
+
+    // chain the operations, applying them with the trace of the next module
+    var operationPromise = Promise.resolve(trace.tree);
+    operations.forEach(function(op) {
+      operationPromise = operationPromise
+      .then(function(curTree) {
+        return builder.trace(op.moduleName)
+        .then(function(nextTrace) {
+
+          var operatorFunction;
+
+          if (op.operator == '+')
+            operatorFunction = builder.addTrees;
+          else if (op.operator == '-')
+            operatorFunction = builder.subtractTrees;
+          else
+            throw 'Unknown operator ' + op.operator;
+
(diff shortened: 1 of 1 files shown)


guybedford:
Awesome. Does `op.moduleName` and `op.operand` make the most sense do you think? Does this system expand to support brackets or order preferences (like the shared tree operator taking preference over addition)? We should make sure our API does expand to these cases.

index.js
```diff
@@ -70,6 +70,38 @@ Builder.prototype.build = function(moduleName, outFile, opts) {
   });
 };
 
+var operatorFunctions = {
+  '+': builder.addTrees,
+  '-': builder.subtractTrees
+};
+
+Builder.buildExpression = function(firstModule, operations, cfg) {
```
guybedford: I think we should make this a class method instead of a static method? Since builds are under performance-pressure, taking advantage of the natural caching of the loader in this way is useful.

crisptrutski: Misunderstood you at the start - used to class/instance distinction (where class vs static is often a quibble about dispatch). No problem being an instance method.

crisptrutski: Doesn't handle rich expressions at all, need some recursion for that. Not sure the value of grouping with just `+` and `-`, unless I'm missing something you can achieve the same effects with reordering and maybe repeating in perverse cases?

New operators coming?

crisptrutski: Not a fan of `op.moduleName`, because there's the implicit connection to the entire previous expression. Having `op.operator` and `op.operands` makes most sense, and we parse `a + b - c` to `{operator: '-', operands: [{operator: '+', operands: ['a', 'b']}, 'c']}`. ASTz

guybedford: There is the potential in future for a few other operators, like an intersection operator (maybe `^`), see https://github.com/jspm/jspm-cli/issues/133 (it's * there). There is also the possibility of weak addition and weak subtraction (adding / subtracting just a single module itself and not its full dependency tree).

guybedford: Also do you think we should include the expression-parsing code here as well?

crisptrutski: Sorry for stunted communication, net keeps dropping here. Yeah I think moving the parse up makes sense. Can imagine gulp chain building or config files containing strings.

guybedford: I'd be inclined towards `operator`, `left` and `right`?

crisptrutski: If you're sure they'll all be binary - what about unary (only think of awful things like, stop at depth 3, or compile with feature flog or altered config for this node) or variadic (`+` works well here)

crisptrutski: Sure, if they're all binary then no reason to allow arrows - just more support for malformed input down the road.

crisptrutski:
Another nit here could be to move the operator function lookup earlier, so errors can be thrown before tracing preceding modules.

guybedford:
Since we've hidden the implementation now, I'm happy to revise as we go.

One thing - did you read my notes re removing configuration arguments in favour of a direct `.config` call. Thoughts?

crisptrutski:
Missed those notes. Yeah that would remove quite a bit of duplication. You're in a rush to hit `0.10.0`? Hitler was right!

guybedford:
Cool, I think otherwise we're just duplicating it in every config call. Absolutely fine with keeping things breaking on this project. The API changes are simple enough to handle the upgrades fine.

crisptrutski:
Just hope SemVer 3.0 comes out before we run out of 32 bit minor versions

Issue #51, opened 2015-01-26
Optimization build functions


guybedford:
Automated two-layer and three-layer bundling into separate files given the a list of modules for each page of the app.

guybedford:
Something like - 

```javascript
var builder = new require('systemjs-builder')(config);

builder.optimize({
  page1: ['page1-module'],
  page2: ['page2-module', 'another-module']
}).then(function(trees, bundles) {
});
```

Bubblyworld:
See [here](https://github.com/Bubblyworld/systemjs-basic-optimize) for a simple optimize function following the above API I wrote for [zygo](https://github.com/Bubblyworld/zygo).

interactionist:
@guybedford When you say three-layer bundling do you refer to have a third list of modules like:
```
  page1: ['page1-module'],
  page2: ['page2-module', 'another-module']
  page3: ['page3-module', 'another-module', 'yet-another-module']
```

So that there is an optimisation problem when you try to find the intersects between the trees of the pages? I am still really keen to get into this. Really try sort out my terminology problem.

guybedford:
@Bubblyworld perhaps you can help explain this area?

Bubblyworld:
@guybedford @interactionist Sure, sorry about the delay!

The way I understand three layered bundling is as follows. As you wrote, you have a list of pages and their traced dependencies. Some dependencies might be unique to a page, some might be shared between a few of the pages and others might be shared by _all_ of the pages.

The idea is to create three layers of bundles:
1. A single common bundle, containing the dependencies shared by all pages.
2. A layer of bundles containing dependencies unique to each page.
3. A layer of bundles containing dependencies shared by some of the pages.

The sample implementation I posted above does this the naive way, by (efficiently) taking every subset of pages and creating a bundle for dependencies shared within each subset. This could potentially result in a large number of bundles, which kinda defeats the point of doing bundling at all. (X_x)

So the problem is to split up the modules into bundles for the best possible user experience. What this means and whether it's even worth solving I'm not sure. I also might be completely missing the mark, so take this with a pinch of salt!

circlingthesun:
Any progress on this? It's the only thing that is stopping me from moving to jspm ;)

interactionist:
@Bubblyworld 
Is the dependency object a flat object like:
```
{
  'page1' : ['module1', 'module2', 'module3'],
  'page2' : ['module1', 'module4']
  'page3' : ['module2', 'module3']
}
```
[...]

PR #82, opened 2015-02-18
Support inline source maps

### LINKED ISSUE #74: Inline source maps option
Write out source map as data URL rather than separate file.

@hgl just creating this issue, as we [basically have the functionality](https://github.com/systemjs/builder/blob/d832c3ef35c264796b86765dadf5503d8a498196/lib/builder.js#L125-L129), but curious to know your use case.

Would you assume / require that the inlined source map always contains `sourcesContent`?

crisptrutski:
Closes #74


lib/builder.js
```diff
@@ -105,14 +105,14 @@ function minify(output, fileName, mangle) {
...
-  if (opts.sourceMaps) {
+  if (output.sourceMaps) {
     var sourceMapFile = path.basename(opts.outFile) + '.map';
     output.source += '\n//# sourceMappingURL=' + sourceMapFile;
   }
 
   return asp(mkdirp)(path.dirname(path.resolve(basePath, opts.outFile)))
   .then(function() {
-    if (!opts.sourceMaps) return;
+    if (!output.SourceMap) return;
```
guybedford: Small `s`?

crisptrutski:
Having some race conditions with `Cannot call method '_nodeRequire' of undefined` when running tests now (via JSX plugin). Settles if running with `--watch`. 

Would love to squish this - something I can block on to know that systemjs is ready?

lib/builder.js
crisptrutski: :open_mouth:

guybedford:
That would be if `global.System` is undefined. This is still needed unfortunately.

crisptrutski:
Travis fails are just that issue, going see about putting in a diabolical spin loop.

guybedford:
I may have fixed this build issue in https://github.com/systemjs/builder/commit/75e8eec59e879e63d67c4cd6b9da65984f0b46bb.

test/expectations/inline-source-map.js
```diff
@@ -0,0 +1,11 @@
...
+
+(function() {
+function define(){};  define.amd = {};
+System.register("tree/amd-2", [], false, function() {
+  return {amd: '2'};
+});
+
+
+})();
+//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRyZWUvYW1kLTIuanMiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6Ijs7Ozs7T0FPTyxFQUFFLEdBQUUsQ0FBRyxJQUFFLENBQUU7RUFBRTtBQUFBIn0=
```
guybedford: I'm worried about all these exact source tests - everytime we update compiler implementations we will be needing to update and generate these tests. How do we update the generated code? etc etc.

crisptrutski: Sure, it's brittle as hell. Big advantage when test coverage is low is that you don't miss details. Updating the generated code could be as easy as setting an environment variable before running the tests again.

Had pretty good experience with "test everything, even the incidental things" as the initial strategy. You can immediately refactor quickly, and it's hard to right targeted tests without refactoring or abusing mocks and spies (which is often more expensive than no tests..)

crisptrutski: Could replace this full-text match with a few assertions and keep high confidence:

1. Only one `//# sourceMappingURL` comment, and it's on the last line
2. Type is `data:application/json;base64`
3. Decoding and parsing it generates a parseable source map

crisptrutski:
@guybedford merge?

package.json
```diff
@@ -22,7 +22,8 @@
     "es6-module-loader": "0.14.0",
     "mocha": "~2.0.0",
     "chai": "^1.10.0",
-    "react-tools": "^0.12.1"
+    "react-tools": "^0.12.1",
+    "atob": "^1.1.2"
```
guybedford: Can we not just use Buffer directly here? https://github.com/node-browser-compat/atob/blob/master/index.js#L5

lib/builder.js
```diff
@@ -105,14 +106,14 @@ function minify(output, fileName, mangle) {
 }
 
 function writeOutputFile(opts, output, basePath) {
-  if (opts.sourceMaps) {
+  if (opts.sourceMap && output.sourceMaps) {
```
guybedford: sourceMaps?

lib/builder.js
```diff
@@ -122,7 +123,7 @@ function writeOutputFile(opts, output, basePath) {
   });
 }
 
-exports.inlineSourceMap = function(output) {
+var inlineSourceMaps = exports.inlineSourceMap = function(output) {
```
guybedford: I generally prefer a pattern here of:

```javascript
exports.inlineSourceMaps = inlineSourceMaps;
function inlineSourceMaps(output) {
}
```

Feels ES6y having declarativey exports.

Also we should try and maintain consistency with the plurals?

index.js
```diff
@@ -24,6 +24,14 @@ function clone(obj) {
   return copy;
 }
 
+function processOpts(opts_, outFile) {
+  var opts = clone(opts_);
+  if (outFile) opts.outFile = outFile;
+  if (opts.sourceMaps == 'inline')
+    opts.sourceMapContents = true;
+  return opts;
+}
```
guybedford: To be really nice to the engine, let's make this function define all the properties on `opts` that need to be used. This way it can be optimized by a single class without hitting deoptimizations later. It also acts nicely as a simple centralisation of info when I forget what all the options do.

crisptrutski: Sure

lib/builder.js
crisptrutski: Good catch. Need tests for `writeOutputFile` evidently.

lib/builder.js
crisptrutski: Don't like that forward reference, but sure the exports do **pop** more this way.

package.json
crisptrutski: Are you sure we want to get on the wrong side of the Modularity Wars?! 

Thanks, will drop `atob`.

lib/builder.js
crisptrutski: Plurals were tricksy - got bitten in both directions. We don't control `output`, but we can change `opts` (queue breaking grunt users again).

guybedford: Can you clarify the cases you'd suggest?

crisptrutski: `opts.sourceMaps` -> `opts.sourceMap`

guybedford: It's a tricky one. We're not a transpiler like Babel that deals in a single source - we work on sources. So I think it makes sense to think in the plural for operations. I think I'm happy with `opts.sourceMaps`, but let me know if you think it's confusing? I was also referring to being consistent with `inlineSourceMaps` and `inlineSourceMap` both being used here.

crisptrutski: Better tests will catch the kinds of slips I made. Missed that typo. Feeling a squash coming on :)

crisptrutski: (Nit: we actually do emit only a single flattened source map, even if we generate intermediate ones. But that'll change with #51)

guybedford: No worries, like being kept on my toes :)

compilers/register.js
```diff
@@ -42,7 +42,7 @@ exports.compile = function(load, opts, loader) {
 
   if (opts.sourceMaps)
     options.sourceMaps = 'memory';
-  if (opts.lowResolutionSourceMaps)
+  if (opts.lowResSourceMaps)
```
guybedford: Nice catch!

lib/builder.js
guybedford: Yes exactly :)

crisptrutski:
Hit all the feedback I think

index.js
```diff
@@ -14,14 +14,25 @@ var builder = require('./lib/builder');
...
+    'normalize',
+    'outFile',
+    'runtime',
+    'sourceMaps',
+    'sourceMapContents'
+  ];
+  var opts = {};
+  keys.forEach(function(key) {
+    opts[key] = opts_ && opts_[key];
+  });
```
guybedford: The point of this is to allow type inference through the setting of default values.

A common pattern I find useful is:

```javascript
function prepend(a, b) {
  for (var p in b) {
    if (!a[p])
      a[p] = b[p];
  }
  return a;
}

function processOpts(opts_, outFile) {
  // apply defaults
  prepend(opts_, {
    config: {},
    lowResSourceMaps: true,
    minify: true,
    normalize: false,
    outFile: '',
    runtime: false,
    sourceMaps: true,
    sourceMapContents: false
  });
}
```

Sorry to be a pain...

guybedford:
Looks great, thanks.

index.js
crisptrutski: Almost added it as `extend(defaults, normalize(opts), overrides)`, but then stopped because really config should also be normalized, and that triggered my subscription to the religion that configuration should never be nested, so then I went back to "smallest thing that meets this nitpickers explicit points" :japanese_ogre: 

Sure we could add this prepend form as a "stopgap" ;)

crisptrutski: Current behaviour has every default or falsey. Assume your example was just to demonstrate, and you're not wanting to enforce new defaults?

guybedford: Ah of course - yes defaults can stay falsey.

crisptrutski:
OK, gave you something similar to your `prepend` case, but without mutating the input argument. Think it's as good enough as we can get without defining a bunch of helpers to give JS basic data wrangling powers :battery: 

(what the hell.. no lambda or wizard emoji when I needed them there)

guybedford:
Nice, thanks! In terms of giving JS basic data wrangling powers, Rick Waldron kindly proposed this for us in ES6 :) The pattern going forward for this in JS will be `Object.assign` (http://www.2ality.com/2014/12/es6-oop.html).

guybedford:
Can we rebase? Then good to merge!

crisptrutski:
Rebased!

Don't link me to articles about OO, I was having such a nice day :cloud:

PR #84, opened 2015-02-18
Upgrade Traceur to 0.0.84

### LINKED ISSUE #79: Update Traceur
Ideally we can update to 0.0.84 and work out the options to set to get source maps working. If there are issues it would be good to follow them up soon so that Traceur doesn't get held back to 0.0.82 for too long.

crisptrutski:
Closes #79

So the change that's messing us around: https://github.com/google/traceur-compiler/commit/728176734acd0accc99132ed64f6bafa628c243b

Would be nice to make this change more elegantly. Maybe bring in a traceur wrapper namespace, that lets us clean up all the duplication option wrangling too.


guybedford:
AMAZING. I think I hear Traceur 0.0.85 in the works so would be nice to jump straight to that.

crisptrutski:
OK will update

crisptrutski:
Ah, "in the works". OK will clean this up quickly then park.

crisptrutski:
Less invasive change, with more wiggle-room for future hassles

guybedford:
Nice, we can probably squeeze in even more of the common code into the compile options.

crisptrutski:
@guybedford so 0.0.85 is out, but breaks tests again since 0.0.84. Looking into it, for one thing the number of mapping is huge, they may have broken lowRes option.

crisptrutski:
@guybedford so the changes look benign, correspond to bug fixes.

 Double checked that changes around low res (changed to strict mode) haven't caused issues too.

RE: easing maintenance of file expectations, added check for `UPDATE_EXPECTATIONS` env variable for now.

crisptrutski:
Looks like the `_nodeRequire` fix has lost its mustard too (strange, didn't see the error for a while and now its very common again)

guybedford:
The only thing here that needs to be manually ported still I believe is the expectations resetting.

I really wanted to like the traceur wrapper idea, and even played around with a few variations of compiler utility, but every encapsulation I attempted was either too big or too small. Traceur's compiler API is really best handled directly I think as there are a lot of little things to watch.

guybedford:
Moved the resetting port to being tracked at https://github.com/systemjs/builder/issues/93.

PR #86, opened 2015-02-20
Support '6to5' with deprecation notice, add warning on unknown

### LINKED ISSUE #85: Builder not transpililng es6?
Hi, first attempt at the builder, so maybe very stupid mistake, but it seems that my module's code is not transpiled from es6 syntax to common-js syntax. (The code works fine just using `System.import('app')`.)

gulpfile.js + systemjs-builder:
```
var path = require('path')
var gulp = require('gulp')
var shell = require('shelljs')
var Builder = require('systemjs-builder');
var builder = new Builder();

var output = 'dist/app-bundle.js'
shell.mkdir('-p', path.dirname(output))

gulp.task('build', function() {    
    builder.reset()
    builder.loadConfig('./src/config.js').then(function() {  

        builder.config({ 
            baseURL: '.' 
        });

        return builder.buildSFX('app', output)
    });
})
```

Loading bundle with
```
<script src="src/jspm_packages/system.js"></script>
<script src="src/config.js"></script>     
<script src="dist/app-bundle.js"></script>   
```

Browser console gives:
```
Uncaught SyntaxError: Unexpected reserved word
app-bundle.js:1382
```

app-bundle.js#L1382-1400
```
import m from 'npm:mithril@0.1.30';
export default {
  controller: function() {
    var name = m.prop("Apex");
    return {name: name};
  },
  view: function(ctrl) {
    return [m("h1", ctrl.name() + " says hi!")];
  }
};

System.register("npm:mithril@0.1.30", ["npm:mithril@0.1.30/mithril"], true, function(require, exports, module) {
  var global = System.global,
      __define = global.define;
  global.define = undefined;
  module.exports = require("npm:mithril@0.1.30/mithril");
  global.define = __define;
  return module.exports;
});
```

crisptrutski:
Closes #85

### DIFF
--- a/compilers/es6.js
+++ b/compilers/es6.js
@@ -34,7 +34,10 @@ exports.compile = function(load, opts, loader) {
 
   var source = load.source;
 
-  if (loader.transpiler == 'babel') {
+  if (loader.transpiler == 'babel' || loader.transpiler == '6to5') {
+    if (loader.transpiler == '6to5')
+      console.log("WARNING: transpiler option should be updated from '6to5' to 'babel'");
+
     options = loader.babelOptions || {};
     options.modules = 'system';
     if (opts.sourceMaps)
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for this!

Unfortunately this isn't going to work all the way down the stack, unless we pre-process the config and change the option. Perhaps just throw if the transpiler is not a valid option (undefined, traceur or babel)? We can include a useful error message if it is 6to5 though.

crisptrutski:
Ah, see what you mean -that transpiler gets around . Agree we should go with exceptions

PR #99, opened 2015-03-04
Organising

guybedford:
Just moved a few things around. @crisptrutski I really hope this doesn't conflict with anything.

I've brought across the Traceur `sourceRoot` change manually, but not the set expectations system.

Moved the builder class into its own file, and made the tree operations static methods.

Also separated out compile functions into a separate file.


guybedford:
The plan tomorrow is to do similar organising to clean up the test folder a bit and then publish a release.

crisptrutski:
Structure definitely better. 

Git failing to grok the renames with amount of changes, so didn't have energy to go through the diffs too much beside where things sit now, and to spot what came through from the Traceur related stuff.

Throwing away rest of #84?

Issue #68, opened 2015-02-10
sourcesContent option


guybedford:
Support inlining all sources into the source map itself.

guybedford:
(Not a high priority currently).

hgl:
Thanks for creating the issue.

Maybe could also provide an option to inline the source map itself in the bundled file with data url?

twalker:
+1

crisptrutski:
@guybedford cost here is low, intermediate source maps (via traceur / 6to5) actually contain `sourcesContent`, we just need to preserve them across concatenation - have got that working, and will put it behind a flag.

There's one gotcha - transitive source maps have the wrong content (after plugin compilation). Anxious this might be a bug upstream, confirming now.

crisptrutski:
OK, the chaining issue is actually related in this case to the source map produced by `react-tools` (no `sourcesContent`). 

In my testing both 6to5 and Traceur are returning source maps listing the jsx file as the source, but with the compiled source content, so this seems to be a bug back in the mozilla library. Will PR back there, we can handle gaps but not lies.

For gaps, I guess we'll just dig back to the file system (to ensure misc plugins can't ruin the parade)

guybedford:
@crisptrutski transitive sourcesContent bubbling up correctly would be ideal and fills gaps as you say. Happy to see this as a nice-to-have feature though, just like the issue here.

crisptrutski:
@twalker as you mentioned on the other thread this being a big deal for you, pushed up a crude solution here https://github.com/systemjs/builder/tree/wip-sources-content if that's useful to you in the mean time.

guybedford:
:+1:

twalker:
Thank you very much @crisptrutski . 

My goal is to transpile with 6to5 and be able to debug ES6 source files. Inlining the source seems less prone to pathing errors.

I'm currently blocked by https://github.com/jspm/jspm-cli/issues/511 so I think I need to roll back instead of forward. I fully appreciate the effort going into getting sourcemaps working though.

guybedford:
Yeah this would definitely be useful for source map portability.

crisptrutski:
Next on my list after auto run-time injection. Likely this weekend.

guybedford:
@crisptrutski can you provide a code example of how I would create a separate source map that inlines all the sources it references?

crisptrutski:
[...]

PR #132, opened 2015-04-09
Pessimistic support for `includeSourcesContent`

crisptrutski:
It's a bit messy - old POC code just rebased. 

Worst part is that it can't rely on existing `sourcesContent` due to (I believe) an issue in Traceur's transitive source maps.


crisptrutski:
Refs #68 (but only marks fixed-pending-release, does not close :sob:)

PR #134, opened 2015-04-13
#51 - Optimization build function

markstickley:
Just wanted to get this in there and under your eyeballs! Let me know what you think.

The optimization function is included in the systemjs-builder package. It is not the most hands-off optimization approach there is but a good one, I think. I can go through it if you like. It's not set as a default if you don't specify a different function - you always have to specify.

There are tests for Builder.prototype.optimizeBuild. The tests for the optimization function are in progress, as are instruction for use in the README... look forward to future pull requests for those!

It currently writes a config.js file to the output directory which needs to be added to the end of the production config. I know this is something currently under review or changing in some way so I didn't spend too much time on that bit. Happy to take suggestions.

Looking forward to hearing any feedback you have. Thanks!


markstickley:
Hmm one of the source maps tests is failing. 3 small differences between expected and actual. I'm not sure what would have caused that; it's fairly independent from other work. Any suggestions would be appreciated, but in the mean time I'll keep digging..!

markstickley:
Weird, I just ran the tests locally and they passed. No changes.

```
systemjs-builder (master)$ npm run test

> systemjs-builder@0.10.4 test /Users/marks/workspace/systemjs-builder
> mocha

undefinedcjs.js
Running in-memory build...
Running a multi-format build...


  Bundle Expressions
    ✓ Addition (935ms)
    ✓ Single module subtraction 
    ✓ Commonality operator (413ms)
    ✓ Wildcard bundling (119ms)
    ✓ Wildcard plugin 

  Builder
    #optimizeBuild
      ✓ should throw if passed bad entryPoints 
      ✓ should throw if passed a bad optimizationFunction 
      ✓ should call the optimizationFunction once (74ms)
      ✓ should pass the optimizationFunction the entry points (48ms)
      ✓ should pass the optimizationFunction a trace object representing the traced entry points (41ms)
      ✓ should pass on the properties of the options object to the optimizationFunction (56ms)
      ✓ should pass an options object with a bundleNameMap property to the optimization function if entryPoints is an Array (44ms)
      ✓ should pass an options object with a bundleNameMap property to the optimization function if entryPoints is NOT an Array 
      ✓ should name the traces based on the entryPoints keys (41ms)
      ✓ should give the traces default names if entryPoints is an Array 
      ✓ should write the bundle files if an outPath is given in the options 
      ✓ should include the source for the bundle files in the returned object if no outPath is given (52ms)
      ✓ should write the config.js if an outPath is given in the options 
      ✓ should resolve its promise with an object containing information about the optimized bundles 
      ✓ should resolve its promise with an object containing information that would allow configuration of System.js to use the bundles 
    #validateOptimizedBundleData
      ✓ should throw if there is a problem with the structure/type of the bundle data 
      ✓ should throw if there is a problem with the structure/type of the bundle data bundles property 
      ✓ should throw if there is a problem with the structure/type of the bundle data config property 
      ✓ should not throw if the structure/type of the bundle data is OK 
      ✓ should allow additional properties to be included in the bundle data 

  Source Maps
    ✓ can render inline 
    sources paths
      ✓ are relative to outFile (175ms)
      ✓ are relative to baseURL, if no outFile (123ms)
    Traceur
      without input source maps
        ✓ handles single compilation targets correctly 
        ✓ handles multiple compilation targets correctly (112ms)
    babel
      without input source maps
        ✓ handles single compilation targets correctly 
        ✓ handles multiple compilation targets correctly (229ms)


  32 passing (3s)
```

lib/builder.js
```diff
@@ -219,6 +220,207 @@ Builder.prototype.traceModule = function(moduleName) {
...
+        sourceMaps: opts.sourceMaps === undefined ? true : !!opts.sourceMaps,
+        mangle: opts.mangle === undefined ? true : !!opts.mangle
+      },
+      function(err) {
+        throw "optimizeBuild: error creating bundle `"+bundleName+"`: "+err;
+      }).
+      then(function(result) {
+        // ...or save the built-tree source as a property on the bundle representation.
+        if(!outPath) {
+          outputBundle.source = result;
```
guybedford: We always set this regardless for `bundle`.

lib/builder.js
```diff
@@ -219,6 +220,207 @@ Builder.prototype.traceModule = function(moduleName) {
...
+        }
+      });
+    })).
+    then(function() {
+      var promise = new Promise(function(resolve, reject) {
```
guybedford: As mentioned, it would be nice to just generate `output.bundles` here.

I think `output.depCache` is a roughly equivalent output anyway, or is it different?

This would allow the user to put the config with the entry points as needed. For example, you probably want to only include the bundles config that applies for specific rendered pages.

lib/builder.js
```diff
@@ -219,6 +220,207 @@ Builder.prototype.traceModule = function(moduleName) {
...
+ *   uglify: true, // optional, defaults to true
+ *   // this option is specific to the optimisation function and may vary between implementations
+ *   entryPointPriorities: {
+ *     main: 0,
+ *     help: 1,
+ *     other: 2
+ *   },
+ * });
+ */
+Builder.prototype.optimizeBuild = function(entryPoints, optimizationFunction, opts) {
```
guybedford: It would be nice to reference all of this in a separate file - `optimize.js` which is separately required.

lib/builder.js
```diff
@@ -247,36 +449,51 @@ function visitTree(tree, moduleName, pluginLoader, visit, seen) {
   });
 }
 
-// returns a new tree containing tree1 n tree2
-Builder.prototype.intersectTrees = function(tree1, tree2) {
-  var name;
-  var intersectTree = {};
+
+// takes 2-N arguments (tree1, tree2, treeN), returns a new tree containing tree1 n tree2 n treeN
+Builder.prototype.intersectTrees = function() {
```
guybedford: Can you rather construct your own intersection reduction function than overloading this one?

lib/builder.js
```diff
@@ -247,36 +449,51 @@ function visitTree(tree, moduleName, pluginLoader, visit, seen) {
...
+
   return intersectTree;
 };
 
-// returns a new tree containing tree1 + tree2
-Builder.prototype.addTrees = function(tree1, tree2) {
-  var name;
-  var unionTree = {};
+// takes 2-N arguments (tree1, tree2, treeN), returns a new tree containing tree1 + tree2 + treeN
+Builder.prototype.addTrees = function() {
```
guybedford: And again.

optimizers/prioritized.js
```diff
@@ -0,0 +1,363 @@
+var path = require("path");
```
guybedford: I do think this whole file should be a separate package actually that is installed and referenced.

guybedford: Then we can separate tests of the actual bundle function from the bundling implementation as well.

guybedford:
@markstickley I've provided some initial feedback here, let me know what you think.

It's nice to just leave this at a framing of the problem and then always install individual implementations - encouraging an ecosystem of optimizations and good tooling around a general bundle optimization interface.

Really appreciated for your work on this.

lib/builder.js
markstickley: No problem... builder.js reference it and include it in the prototype or should optimizer be completely independent from the Builder object?

lib/builder.js
markstickley: OK that's fine, I'll remote the `if`

lib/builder.js
markstickley: I was under the impression that the `bundles` config was keyed on the name of each bundle and contained a list of all files in the bundle while the `depCache` config was keyed on individual files and contained a list of dependencies for that file. I suppose that `depCache` would be redundant providing your optimisation function included all dependencies for an endpoint in each bundle, but that's not necessarily going to be the case always...

Happy to remove the bit that writes the config file, I wasn't really sure where to put it anyway.

lib/builder.js
markstickley: I certainly can, although it shouldn't affect the function of it, or event impact the efficiency in any great way.

optimizers/prioritized.js
markstickley: OK no problem. Are you thinking optimization packages/functions should be hosted within the systemjs ecosystem or just on their creators' accounts?

markstickley:
No problem, sorry for taking so long to come back to you on this - I got put on another project but I have a bit of slack time now so I'm looking at it again.

I've replied to your inline comments, seeking a little clarification in some. It would be great to get a version of this merged! Happy to make the necessary changes...

PR #140, opened 2015-04-14
exposing original config settings object to users of instances of Builder

capaj:
I couldn't find the original setting anywhere on the builder object. If they are available for access even without this change, please tell me how to access them and close this PR.
If not, please merge.


guybedford:
@capaj the config can be read from `builder.loader.paths`, `builder.loader.map` etc. Let me know if that is sufficient?

capaj:
@guybedford  thanks, that works for me.

PR #142, opened 2015-04-16
Enable opts as second arg in build, buildTree & buildSFX

origin1tech:
Check if second arg is options instead of outFile. added "isPlainObject" helper. Falls back if Object.getPrototypeOf is not defined.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -30,6 +30,12 @@ function processOpts(opts_, outFile) {
   return opts;
 }
 
+function isPlainObject(obj) {
+  if(Object.getPrototypeOf)
+    return typeof obj === 'object' && Object.prototype === Object.getPrototypeOf(obj);
+  return typeof obj === 'object' && Object === obj.constructor;
+}
+
 function Builder(cfg) {
   this.loader = null;
   this.reset();
@@ -121,6 +127,14 @@ Builder.prototype.config = function(config) {
 
 Builder.prototype.build = function(expression, outFile, opts) {
   var self = this;
+
+  // allow passing opts
+  // as second argument.
+  if(isPlainObject(outFile)){
+    opts = outFile;
+    outFile = undefined;
+  }
+
   opts = opts || {};
 
   if (opts.config)
@@ -141,6 +155,14 @@ function addExtraOutputs(output, tree, opts) {
 Builder.prototype.buildTree = function(tree, outFile, opts) {
   var loader = this.loader;
   var self = this;
+
+  // allow passing opts
+  // as second argument.
+  if(isPlainObject(outFile)){
+    opts = outFile;
+    outFile = undefined;
+  }
+
   opts = processOpts(opts, outFile);
 
   return compileOutputs(loader, tree, opts, false)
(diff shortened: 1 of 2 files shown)


lib/builder.js
```diff
@@ -30,6 +30,12 @@ function processOpts(opts_, outFile) {
   return opts;
 }
 
+function isPlainObject(obj) {
+  if(Object.getPrototypeOf)
+    return typeof obj === 'object' && Object.prototype === Object.getPrototypeOf(obj);
+  return typeof obj === 'object' && Object === obj.constructor;
+}
```
guybedford: This is failing in the tests.

lib/builder.js
```diff
@@ -141,6 +155,14 @@ function addExtraOutputs(output, tree, opts) {
 Builder.prototype.buildTree = function(tree, outFile, opts) {
   var loader = this.loader;
   var self = this;
+
+  // allow passing opts
+  // as second argument.
+  if(isPlainObject(outFile)){
```
guybedford: `typeof outFile === 'object'` would be fine here.

crisptrutski:
@origin1tech kills build, since `typeof null === 'object'` but ` Object.getPrototypeOf` cannot be called on `null` (**wat**)

@guybedford slightly uncomfortable with type based approach dispatch here - a very fancy way to essentially sugar omitting the second parameter.

Would favour switching argument order in the 0.17 release instead, then `outFile` can simply be omitted. Is it common to instantiate a `Builder` without passing `opts`?

guybedford:
Thanks, I've commented in line.

lib/builder.js
origin1tech: forgot about null check, makes sense I can add that to the helper func or as you mentioned just do a simple typeof object check which do u prefer?

Regards,

Carey Hazelton
chazelton@origin1.com <mailto:chazelton@origin1.com>
Confidentiality Notice: The information contained in this electronic transmission, is privileged and confidential and is intended only for the recipient(s) named above. If the reader of this message is not the recipient(s) named above, or an authorized agent of such recipient(s) responsible for delivering it to the intended recipient(s), you are hereby notified that you have received this electronic transmission in error. Any review, dissemination, distribution, or copying of this electronic transmission including any attachments is strictly prohibited. If you have received this electronic transmission in error, please notify the sender immediately. Origin1 Technologies.

crisptrutski: I don't believe that would be correct - when callingl `new Builder(tree, null, {"real": "options})` that would take `opts` as `{}`.

origin1tech: disclosure just getting feet wet with builder but seems like new instance of Builder accepts a cfg which is of type string (loads config) or an actual config object. Am I missing something?

Regards,

Carey Hazelton
chazelton@origin1.com <mailto:chazelton@origin1.com>
Confidentiality Notice: The information contained in this electronic transmission, is privileged and confidential and is intended only for the recipient(s) named above. If the reader of this message is not the recipient(s) named above, or an authorized agent of such recipient(s) responsible for delivering it to the intended recipient(s), you are hereby notified that you have received this electronic transmission in error. Any review, dissemination, distribution, or copying of this electronic transmission including any attachments is strictly prohibited. If you have received this electronic transmission in error, please notify the sender immediately. Origin1 Technologies.

PR #143, opened 2015-04-16
Enable opts as second arg in build, buildTree & buildSFX

crisptrutski:
This is just #142 with a bugfix

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -16,11 +16,11 @@ function processOpts(opts_, outFile) {
     config: {},
     outFile: outFile,
     normalize: true,
-    
+
     runtime: false,
     minify: false,
     mangle: true,
-    
+
     sourceMaps: false,
     sourceMapContents: opts_ && opts_.sourceMaps == 'inline',
     lowResSourceMaps: false,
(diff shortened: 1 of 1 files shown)


lib/builder.js
```diff
@@ -30,6 +30,13 @@ function processOpts(opts_, outFile) {
   return opts;
 }
 
+function isPlainObject(obj) {
```
guybedford: Not sure I understand the value of all this still.

lib/builder.js
```diff
@@ -141,6 +156,14 @@ function addExtraOutputs(output, tree, opts) {
 Builder.prototype.buildTree = function(tree, outFile, opts) {
   var loader = this.loader;
   var self = this;
+
+  // allow passing opts
+  // as second argument.
+  if(isPlainObject(outFile)){
```
guybedford: if (typeof outFile === 'object' && outFile !== null)

crisptrutski: Even better `if (outFile && typeof outFile === 'object')` IMHO (less characters and faster). Yeah I'm happy with that, don't see why we care about prototype. An "actual" `outFile` parameter certainly wouldn't be a class instance.

crisptrutski:
Oh man, broke it again. Thumbelina

guybedford:
Beautiful. Squash and merge?

crisptrutski:
Just wanted to remove the my commit pun? :sob:

PR #153, opened 2015-04-24
adding Object.prototype.hasOwnProperty.call for IE8

AutoSponge:
This is a stop-gap measure needed for a production release supporting IE8 prior to the next coordinated release of jspm, system, and es6-module-loader.


guybedford:
Nice! Well this is better than the approach in SystemJS... will port this one back there then.

guybedford:
Thanks for this - I completely missed we could use hasOwnProperty in IE8. Included this in https://github.com/systemjs/builder/commit/d019b8e9ff1dbaa26e2d6bc081ebb857e08fd9e8 with a little more refactoring. Release to follow shortly.

Issue #157, opened 2015-04-27
Check sourceMap doesn't contain file: URLs


guybedford:
@lookfirst mentioned `file:` URLs were turning up int he source map, we should check that isn't from builder here.

lookfirst:
Thanks. Yes, it is from the [systemjs-route-bundler](https://github.com/Swimlane/systemjs-route-bundler) which has been updated to use the latest systemjs/builder code now. 

I disabled it with sourceMaps: false.

crisptrutski:
Bit swamped at the moment, but will try look into this tomorrow

crisptrutski:
@lookfirst thanks for linking me to that test suite, made my life much easier :)

Looked into this, and was able to resolve with some changes to both `builder` and `systemjs-route-bundler`. The issue basically boils down to absolute vs relative paths (`builder` assumes `baseURL` is an absolute path), and the informal `file:/` specification (which node's `path` does not understand, but `builder` assumes `baseURL` is using).

@guybedford  

Can send a PR for `builder` to be more agnostic around absolute file paths (in `sources` and `baseURL`), and to reject or at least not treat-as-if-it-were-absolute relative paths.

Don't think that `builder` should be deciding how to interpret relative paths though, in the context of normalizing the absolute paths that come back from loaders. Using `process.cwd()` matching "first occurance" or "last occurance" of the relative path in loaded file paths would work in many cases, but seem too opinionated.

In my testing I just modified the route-bundler to expand the path before calling builder, [just like jspm does](https://github.com/jspm/jspm-cli/blob/cd751a5408e40f5f134c613a246a462f891d643f/lib/bundle.js#L37), and believe this is the correct solution.

PR #158, opened 2015-04-30
URL hygiene

crisptrutski:
@guybedford 

This cleans up sticky tape around `file:`, which could go horribly wrong (see #157) and for example rebase absolute URLs against a relative baseURL.

Presenting two approaches: strict (see after first commit) and permissive (see after last commit)

If we decide to go with either of these I'll add some tests.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -4,6 +4,8 @@ var fs = require('fs');
 var Promise = require('rsvp').Promise;
 var asp = require('rsvp').denodeify;
 
+var coercePath = require('./utils').coercePath;
+
 function countLines(str) {
   return str.split(/\r\n|\r|\n/).length;
 }
@@ -53,17 +55,17 @@ function processOutputs(outputs) {
   return outputObj;
 }
 
-function createOutput(outputs, outFile, baseURL, opts) {
+function createOutput(outputs, outFile, outputPath, opts) {
   // process output
   var saucy = require('./sourcemaps');
   var sourceMap;
+  var outputURL = 'file:/' + outputPath;
 
   var outputObj = processOutputs(outputs);
 
   if (opts.sourceMaps && outputObj.sourceMapsWithOffsets.length) {
-    var sourceRoot = outFile ? path.dirname(baseURL + outFile) + path.sep : baseURL;
     var mapsWithOffsets = outputObj.sourceMapsWithOffsets;
-    sourceMap = saucy.concatenateSourceMaps(outFile, mapsWithOffsets, sourceRoot, opts.sourceMapContents);
+    sourceMap = saucy.concatenateSourceMaps(outFile, mapsWithOffsets, outputURL, opts.sourceMapContents);
   }
 
   var output = outputObj.sources.join('\n');
@@ -81,7 +83,7 @@ function minify(output, fileName, mangle, globalDefs) {
   ast = ast.transform(uglify.Compressor({
     dead_code: true,
     global_defs: globalDefs,
-    warnings: false 
+    warnings: false
   }));
   ast.figure_out_scope();
   ast.compute_char_frequency();
@@ -139,13 +141,17 @@ function inlineSourceMap (output) {
 }
 
 exports.writeOutputs = function(opts, outputs, baseURL) {
-  // remove 'file:' part
-  var basePath = baseURL.substr(5);
+  var basePath = coercePath(baseURL);
+  var outputPath;
 
-  if (opts.outFile)
+  if (opts.outFile) {
     opts.outFile = path.relative(basePath, path.resolve(opts.outFile));
+    outputPath = path.dirname(path.resolve(basePath, opts.outFile));
+  }
+  else
+    outputPath = basePath;
 
-  var output = createOutput(outputs, opts.outFile, baseURL, opts);
+  var output = createOutput(outputs, opts.outFile, outputPath, opts);
 
   if (opts.minify)
     output = minify(output, opts.outFile, opts.mangle, opts.globalDefs);
--- a/lib/output.js
+++ b/lib/output.js
@@ -59,7 +59,8 @@ function createOutput(outputs, outFile, outputPath, opts) {
   // process output
   var saucy = require('./sourcemaps');
   var sourceMap;
-  var outputURL = 'file:/' + outputPath;
+  var onWindows =  process.platform.match(/^win/);
+  var outputURL = (onWindows ? 'file://' : 'file:/') + outputPath;
 
   var outputObj = processOutputs(outputs);
 
--- a/lib/sourcemaps.js
+++ b/lib/sourcemaps.js
@@ -2,6 +2,8 @@ var sourceMap = require('source-map');
 var traceur = require('traceur');
 var path = require('path');
 var fs = require('fs');
+var filePath = require('./utils').filePath;
+var isFileURL = require('./utils').isFileURL;
 
 var wrapSourceMap = function(map) {
   return new sourceMap.SourceMapConsumer(map);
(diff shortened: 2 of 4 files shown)


lib/output.js
```diff
@@ -139,13 +140,19 @@ function inlineSourceMap (output) {
...
 exports.writeOutputs = function(opts, outputs, baseURL) {
-  // remove 'file:' part
-  var basePath = baseURL.substr(5);
+  var basePath = filePath(baseURL);
+  var outputPath;
 
-  if (opts.outFile)
+  if (opts.outFile && basePath) {
     opts.outFile = path.relative(basePath, path.resolve(opts.outFile));
+    outputPath = path.dirname(path.resolve(basePath, opts.outFile)) + path.sep;
```
guybedford: Does this need a trailing '/'? If so are we guaranteeing it for filePath?

crisptrutski: Just cargo culting the separator, while check if its required

lib/utils.js
```diff
@@ -0,0 +1,11 @@
+var path = require('path');
+
+/* Normalize URL or path to an absolute path, if it corresponds to one. */
+function filePath(urlOrPath) {
+  // give naked paths the benefit of the doubt that they are file paths
+  if (path.isAbsolute(urlOrPath))
```
guybedford: I'm not sure `http://some-site.com/source.js` would pass this?

lib/sourcemaps.js
```diff
@@ -71,14 +72,16 @@ exports.concatenateSourceMaps = function(sourceFilename, mapsWithOffsets, source
 
   // convert from library internals format to canonical
   var normalized = JSON.parse(JSON.stringify(generated));
+  var sourcePath;
 
   if (includeSourcesContent) {
     normalized.sourcesContent = normalized.sources.map(function(source) {
```
guybedford: Does this `source` take into account the `sourceRoot` already?

lib/utils.js
```diff
@@ -0,0 +1,11 @@
+var path = require('path');
+
+/* Normalize URL or path to an absolute path, if it corresponds to one. */
+function filePath(urlOrPath) {
```
guybedford: I'm not sure any tool would output absolute file system paths by default?

If a source URL looked like `/source.js` then it is probably implied that it is rootURL relative.

lib/output.js
crisptrutski: Don't see any reason to add the trailing separator - only used with `path.relative`, which doesn't care.

lib/utils.js
crisptrutski: Non `file:/` URLs are meant to return `null` here. Is the docstring unclear?

crisptrutski: "path" synonymous with "filesystem path", probably misleading

lib/sourcemaps.js
crisptrutski: Depends on the compilation path. Generally these are absolute URLs from Traceur, but otherwise no assumptions are made. 

This function just overrides existing `sourceContents` with actual file contents, if the source has an absolute path. Not assuming relative paths are relative to the `sourceRoot`, because that's the output path of the related source file, which transpilers and plugins are not aware of.

File contents taking precedence to included source is just a workaround for the Tracuer bug around `sourcesContent`.

lib/utils.js
guybedford: `path.isAbsolute('http://site.com') === false`, `path` is specifically for file-system paths. Perhaps you meant to use URL?

lib/utils.js
crisptrutski: Good point. I'll go back and look at where "absolute relative" paths should be sanitized

lib/utils.js
crisptrutski: Yeah the `path.isAbsolute('http://site.com') === false` is assumed here - the purpose of the function is to return a path, if the input looks like a path, or is a file URL. `'http://site.com` is neither, so it would fail both `if`s and return something falsey (`undefined`).

guybedford: I see, thanks for clarifying.

crisptrutski:
@guybedford possibly not to your style liking, but I think this is behaviour we can agree on. Let me know if you'd like to squash the meandering history, or inline / restyle anything.

lib/output.js
```diff
@@ -53,17 +55,17 @@ function processOutputs(outputs) {
   return outputObj;
 }
 
-function createOutput(outputs, outFile, baseURL, opts) {
+function createOutput(outputs, outFile, outputPath, opts) {
   // process output
   var saucy = require('./sourcemaps');
   var sourceMap;
+  var outputURL = 'file:/' + outputPath;
```
guybedford: In Windows, we need to manage adding an extra `/`.

lib/utils.js
```diff
@@ -0,0 +1,25 @@
...
+function filePath(url) {
+  if (isFileURL(url))
+    return url.replace(/^file:\/+/, '/');
+}
+exports.filePath = filePath;
+
+/* Coerce URLs to paths, assuming they are file URLs */
+function coercePath(url) {
+  // give absolute URLs benefit of the doubt that they are file paths
+  if (path.isAbsolute(url))
```
guybedford: This case will never make sense, perhaps we should just skip this.

lib/sourcemaps.js
```diff
@@ -71,27 +73,28 @@ exports.concatenateSourceMaps = function(sourceFilename, mapsWithOffsets, source
...
     //if (contentsBySource[source])
       //return contentsBySource[source];
 
-    if (source.substr(0, 6) === 'file:/')
-      return fs.readFileSync(source.replace(/^file:\/+/, '/')).toString();
+    sourcePath = filePath(source);
```
guybedford: We should probably handle relative sources here? Ie, if it starts with `x.js` or `./x.js` then it should be taken to be relative to sourceFileName?

lib/sourcemaps.js
```diff
@@ -71,27 +73,28 @@ exports.concatenateSourceMaps = function(sourceFilename, mapsWithOffsets, source
...
+    if (sourcePath)
+      return fs.readFileSync(sourcePath).toString();
```
guybedford: Don't we need to wrap this for failure?

lib/sourcemaps.js
```diff
@@ -71,27 +73,28 @@ exports.concatenateSourceMaps = function(sourceFilename, mapsWithOffsets, source
...
 
     // remove this is optimistic return is reliable
     return contentsBySource[source];
     });
   }
 
-  // convert paths to relative
+  var rootIsPath = isFileURL(sourceRoot);
```
guybedford: Can we not make it an assumption of the function that the sourceRoot input must be in the URL space? I'm happy to forgo hygiene when we're within safe internal APIs.

guybedford:
Ok feedbacked, mostly looks great to me, just trying to hit the questions while we're here.

lib/sourcemaps.js
crisptrutski: Makes sense

lib/utils.js
crisptrutski: Would argue that it makes as much sense as relating relative paths to process directory - it's just consistency. We actually get this behaviour for-free from that case though (at the cost of calling `process.cwd()` despite its value having no effect on the result) - so happy to remove this branch.

lib/output.js
crisptrutski: Using `process.platform.match(/^win/)` to determine?

lib/sourcemaps.js
crisptrutski: Yes we should.

lib/sourcemaps.js
crisptrutski: The intention here is to support a non-file URL, rather than a path. Can drop the non-used flexibility.

lib/output.js
guybedford: Yes exactly.

crisptrutski:
- [ ] testlibs
- [ ] manual builder tests
- [ ] squash
- [ ] profit

crisptrutski:
@guybedford ping-a-ling, think this is good to go

lib/output.js
```diff
@@ -53,17 +55,18 @@ function processOutputs(outputs) {
   return outputObj;
 }
 
-function createOutput(outputs, outFile, baseURL, opts) {
+function createOutput(outputs, outFile, outputPath, opts) {
   // process output
   var saucy = require('./sourcemaps');
   var sourceMap;
+  var onWindows =  process.platform.match(/^win/);
+  var outputURL = (onWindows ? 'file://' : 'file:/') + outputPath;
```
guybedford: To be super difficult, we're still in a world of `file://` for now... That is, one less slash!

guybedford:
:+1:

lib/output.js
crisptrutski: You mean remove the windows condition?

PR #161, opened 2015-05-05
Update README to describe bundleSFX arithmetic

nkoterba:
Added documentation to warn users that SFX bundles can only use addition and wildcard arithmetic.


guybedford:
Sure, thanks.

Issue #168, opened 2015-05-12
AMD named define variations


guybedford:
From https://github.com/systemjs/systemjs/issues/446:

`define('name', [], {})` should be supported in builds.

Issue #167, opened 2015-05-11
AMD define replacement is not always reliable


guybedford:
From https://github.com/jspm/jspm-cli/issues/732#issuecomment-100714216, in Kendo UI we get:

```javascript
(function (f, define) {
    define([], f)
})(function () { 
      <<SOURCE CODE HERE>> 
}, typeof define == "function" && define.amd ? define : function (_, f) {
    f()
});
```

Turned into

```javascript
(function() {
function define(){};  define.amd = {};
(function(f, define) {  <-- Actual start of library source
  define([], f);
})(function() {
  <<SOURCE CODE HERE>>
}, typeof define == "function" && define.amd ? define : function(_, f) {
  f();
}); <-- Actual end of library source
})();
```

Because the define function is passed as a function parameter, the AMD transformer doesn't detect this pass to be able to replace the call of define with `System.register`.

It may be preferable to alter the AMD conversion to rather include the `System.register` call within the dummy define function at the top rather so we can handle these cases.

nkoterba:
@guybedford :+1: 

I think the idea of replacing the `define` dummy function at the top would be a nice overall addition if it doesn't break other constructs.

It was confusing initially to see the following in the "bundled" source:
```javascript
function define(){};  define.amd = {};
```

Eventually, I realized that this was "tricking" the AMD module into thinking it was within an AMD environment and to behave like an AMD "module".

In addition, by placing the System.register call at the start of the library's source code it would be easier to always locate.  Right now the `System.register` replacement happens in the 3rd party library wherever they happen to use the `define` function which can be at the start of the file, the end of the file, etc.

guybedford:
Another case of something similar also came up from https://github.com/jspm/jspm-cli/issues/750:

```javascript
  System.register("npm:node-forge@0.6.26/js/forge", [...], false, function(__require, __exports, __module) {
    return (function() {
      defineFunc.apply(null, Array.prototype.slice.call(arguments, 0));
    }).call(this, __require, __module, ...);
  });
```

Where `arguments` loses the deps context throwing in the bundle.

codylindley:
Guy. I'm confused. Did you fix this issue or just decide it was not a bug anymore. With kendo UI, in AMD format the bundle is still breaking. I'd like to really understand the correct fix here.

guybedford:
@codylindley as mentioned, this bug was fixed in https://github.com/systemjs/builder/issues/228. Working on getting the next beta out really soon, just need to smooth out quite a few things still.

[...]

Issue #164, opened 2015-05-07
AMD contextual require support


guybedford:
When the contextual require is present in an AMD module, and it has a second callback argument or an array first argument, then we need to wrap the require function with the following:

```javascript
  function require(names, callback, errback) {
    if (typeof names == 'string') {
      if (typeof callback != 'function')
        names = [names];
      else
        return __require(names);
    }
    if (names instanceof Array) {
      var dynamicRequires = [];
      for (var i = 0; i < names.length; i++)
        dynamicRequires.push(System['import'](names[i], __module.id);
      Promise.all(dynamicRequires).then(callback, errback);
    }
  }
```

PR #170, opened 2015-05-14
AMD Refactoring

guybedford:
AMD layer refactoring to support https://github.com/systemjs/builder/issues/168, https://github.com/systemjs/builder/issues/167, https://github.com/systemjs/builder/issues/164.


guybedford:
- [ ] Bundle tracing now needs to be aware of AMD named defines. Internal named defines within folds will be an issue here. Perhaps there is some way to replace the define name with a string variable that is passed through via a System.register call after all.
- [ ] More testing is needed against as many real-world cases as possible, as this has the potential to break things if we don't get enough coverage
- [ ] SFX template is needed for the new AMD helpers

guybedford:
Merged into 0.11 branch.

PR #185, opened 2015-06-05
SFX module format

guybedford:
This is just a very first sketch, I haven't tested or anything. Just a placeholder for the final work.


theefer:
That looks promising!

PR #189, opened 2015-06-06
AMD Bundle Format Test

nnnnathann:
Hi,

AMD bundling would be really great for me, I created this initial test case but would like some feedback on how to proceed...

Not sure how you want it to work with multiple exports?  I have designed the wrapper right now to publish the contents of the export definitions rather than the manifest in the case that it only has one output key...

Thanks for the great work, hope I can help somehow?

nathan

### DIFF
--- a/templates/sfx-amd.js
+++ b/templates/sfx-amd.js
@@ -1,6 +1,15 @@
 (function(factory) {
-  if (typeof define == 'function' && define.amd)
-    define(factory);
-  else
+  if (typeof define == 'function' && define.amd){
+    define(function () {
+      var obj = factory.apply(this, arguments);
+      var keys = Object.keys(obj);
+      if (keys.length === 1) {
+        return obj[keys[0]];
+      } else {
+        return obj;
+      }
+    });
+  } else {
     factory();
+  }
 })
--- a/test/bundle-format.js
+++ b/test/bundle-format.js
@@ -1,0 +1,16 @@
+var Builder = require('../index');
+var builder = new Builder('test/fixtures/test-tree');
+var spawn = require('child_process').spawn;
+
+suite('Bundle Format', function() {
+  test('AMD', function(done) {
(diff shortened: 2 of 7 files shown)


templates/sfx-amd.js
```diff
@@ -1,6 +1,15 @@
 (function(factory) {
-  if (typeof define == 'function' && define.amd)
-    define(factory);
-  else
+  if (typeof define == 'function' && define.amd){
+    define(function () {
+      var obj = factory.apply(this, arguments);
+      var keys = Object.keys(obj);
```
guybedford: I was planning on just having the first module from the list of modules in the sfx bundle as the exporting module. I think that's a valid tradeoff and keeps things simple. (If I had my way, SFX bundles wouldn't allow addition actually)

test/bundle-format.js
```diff
@@ -0,0 +1,16 @@
+var Builder = require('../index');
+var builder = new Builder('test/fixtures/test-tree');
+var spawn = require('child_process').spawn;
+
+suite('Bundle Format', function() {
+  test('AMD', function(done) {
+    builder.buildSFX('sfx-format-01.js', 'test/output/sfx-amd.js', { sfxFormat: 'amd' })
```
guybedford: Nice!

guybedford:
This looks great! Thanks so much for working on this, it was just needed. I think we're good to merge pending the discussion re handling exports from the first module only.

guybedford:
Merged manually - thanks so much for this!

nnnnathann:
No problem.  Thanks for the excellent work on this library.

guybedford:
@nnnnathann I've removed the multiple exports functionality as discussed above - it just exports the first module regardless. In cases where the other modules need some kind of handle, rather just have a main wrapper module that brings everything together.

nnnnathann:
Cool, sounds good

Issue #163, opened 2015-05-05
Allow bundles themselves to be supported in addition


guybedford:
And avoid this error - 

TypeError: "init" was defined via a bundle, so can only be used for subtraction or union operations.

ohjames:
Thanks so much for opening this ticket.

I've just finished a new asset pipeline and I want to bring incremental building using multiple CPUs to my `jspm` work-flow.

Essentially in this system I pre-compile the assets server side as the project is *massive* with many many `babel` dependencies. Having to wait on each load for babel to compile them all causes big delays to testing.

So instead I compile only the files that change using a build system that watches all the source files meaning the act of compiling ES6 modules is only performed once and in parallel.

With this support I can use ES6 modules all the way, currently I have to use intermediary AMD compiled resources and these come with many disadvantages (e.g. having to maintain dependency order when many modules are bundled into a single file).

ohjames:
Hey @guybedford I'd like to work on this issue and submit a PR, would that be okay? If you have the time would be great if you had any pointers.

With this issue out of the way I can bring incremental rebuilds of bundles to jspm removing the one pain point from jspm that prevents people from migrating to jspm from webpack/browserify.

guybedford:
Thanks so much for getting in touch - it would be great to have a contribution here. I think it may make sense to approach this at an even lower level, by separating compile into _compile load_ and _concat bundle_.

Currently both of the above steps happen in https://github.com/systemjs/builder/blob/master/lib/compile.js#L40.

By separating this into two parts, it would allow a caching system to just make the compile calls it needs using the compiled `source`, `sourceMap` entries (output of compileLoad at https://github.com/systemjs/builder/blob/master/lib/compile.js#L74) already from cache, and then concatenating fresh compiles along with cached compiles together.

Just doing this separation of writing `exports.compileOutputs` as a process calling `exports.compileLoad` and `exports.concatOutputs` would allow a custom build tool like yours to call into `require('systemjs-builder/lib/compile').compileLoad` etc directly in creating a cache-abstracted build tool.

That's the simple low-level API I can think of offhand, in turn it may make sense to add cache abstractions but maybe even not, and to leave that to higher level tools like a jspm watch server etc.

Just let me know if you have any more questions.

guybedford:
Note we should probably export these methods on the builder class instance itself though - `builder.compileLoad` and `builder.concatOutputs`

ohjames:
[...]

PR #193, opened 2015-06-10
Add optional compiler cache to builder.buildTree

ohjames:
As you suggested in https://github.com/systemjs/builder/issues/163 took a while to implement as Splatoon was really good.

Would be good to get some feedback. The test commands weren't working for me, you can see I tried to fix it in the second commit. The pages loaded in the browser didn't seem to be working though (with the untouched master), unless I'm supposed to be looking at one blank page and one blank page with a title. I made sure the `test/output/sfx.js` didn't differ from the unmodified code though.


ohjames:
Had to add a few extra parameters to `compileLoad`, they're only necessary when `sfxEntryPoints` is used, I added some jsdoc to the builder class to explain them.

ohjames:
My 3am version of this code is much better than my 2am version ;)

guybedford:
Yes the tests do need some love. Thanks for the attempt here!

* sfxEntryPoints and plugins don't need to be inputs to compileLoad
* Also compileLoad only creates a single output object, so we shouldn't need to loop and form any ouputs array, it should just be as simple as looking up the format and using that compiler.

ohjames:
Thanks for getting back to me so quickly! Have a few questions for you so that I can finish this off.

> sfxEntryPoints and plugins don't need to be inputs to compileLoad

* For `sfxEntryPoints`: How to deal with the `sfxEntryPoints` + `load.metadata.loader + load.metadata.build === false` case? Seems we get a different output in that case.
* For `plugins` don't we need to keep track of `load.metadata.loaderModule` for modules that use it for eventual passing to `concatOutputs`?

I originally had the code for this inside of `compileOutputs` but I wasn't sure if users of `compileLoad` would need to use this behaviour.

> Also compileLoad only creates a single output object...

Ah of course, fixing it now will only take a minute.

ohjames:
Okay I fixed the `compileLoad` API with respect to multiple outputs, will clean up the `sfxEntryPoints`/`plugins` usage as soon as I have your advice on how to deal with that. Thanks again!

guybedford:
Great, thanks! The important information from the compile that is needed is the module format which informs the SFX system helpers, and the plugin instance which informs the plugin bundling. The module format should be fine to read from the load record though when concatting the outputs. The plugin instance may be the only thing necessary to return separately. Perhaps output.plugin or something could handle this here since it is a single output?

guybedford:
Alternatively we could even load the plugin entirely separately again in the output function. It may be a better separation.

lib/compile.js
```diff
@@ -13,14 +13,28 @@ var compilerMap = {
...
 
 exports.compileLoad = compileLoad;
-function compileLoad(loader, load, opts, compilers) {
+function compileLoad(loader, load, opts, plugins, sfxEntryPoints, compilers) {
+  if (! load.metadata.build) {
+    if (sfxEntryPoints && load.metadata.loader)
+      return 'System.registerDynamic("' + load.name + '", [], false, function() { console.log("SystemJS Builder - Plugin for ' + load.name + ' does not support sfx builds"); });\n';
+    else
+      return null;
+  }
```
guybedford: This should be moved out to before calling compileLoad in compileOutputs.

ohjames:
Thanks for the advice, I see where you're going with this, it sounds much cleaner than what I implemented. Will work on it this evening, should have something for you to review by tomorrow morning.

lib/compile.js
```diff
@@ -13,14 +13,28 @@ var compilerMap = {
...
+
+  // support plugin "bundle" reduction hook
+  var plugin = load.metadata.loaderModule;
+  if (plugin) {
+    var entry = plugins[load.metadata.loader] = plugins[load.metadata.loader] || {
+      loads: [],
+      bundle: plugin.bundle
+    };
+    entry.loads.push(load);
+  }
```
guybedford: This can be moved to concatOutputs

lib/compile.js
```diff
@@ -13,14 +13,28 @@ var compilerMap = {
...
   .then(function() {
     var format = load.metadata.format;
-    if (load.metadata.build === false) {
-      return {};
-    }
-    else if (format in compilerMap) {
+    if (format in compilerMap) {
       // note which compilers we used
       if (compilers)
         compilers[format] = true;
```
guybedford: The compiler tracking can also happen in compileOutputs.

guybedford:
Thanks, just trying to get a strong separation of concerns here.

ohjames:
I actually had a much simpler `compileLoad` in an earlier version of my code, closer to how your code review comments say it should be now, then I started worrying about consumers of the API but I didn't think about how easy it would be to move that functionality into `concatOutputs`, your advice is so helpful :+1:

ohjames:
I'll work on improving the API in a single commit so you can see what's changed, then I'll merge the commits back together when you're happy with the code using `git rebase -i` so you can have some clean history to merge.

ohjames:
Thanks so much for helping me with this, I've made the API so much cleaner now. Let me know if this looks good and I can merge the commits together ready to pull!

Thanks again and sorry for the delay, my keyboard broke and the replacement I ordered was faulty :(

ohjames:
Maybe instead of having to pass both `tree` and `outputs` to `concatOutputs`, `compileLoad` could attach `output` to the `load` (still returning a promise that resolves to the output). Then you'd only need to keep track on one `tree` object for the API. A plugin could simply `RSVP.hash`/`Promise.props` over the module tree to get the output needed for `concatOutputs` then.

What do you think?

ohjames:
Having another look at this now the tests have landed.

ohjames:
So great to see those tests, they caught a serious bug with the code! Amended the commit now to latest master with all tests passing :+1:

Gonna add the `outputs` change I discussed in another commit unless I hear otherwise, will be easy to revert anyway.

ohjames:
I've merged the commits ready for pull. Wrote a plugin for [sigh](http://github.com/sighjs/sigh) that uses this API, get incrementally rebuilt bundles in less than 100ms on a massive Aurelia project, this shaves *so much time* off of our development process (test process gone from 28 seconds to 6 seconds). Looking forward to it being officially released!

guybedford:
Thanks for your work on this. It will be really nice to get incremental builds going.

I know this is the API I suggested, but I'm still not feeling completely happy with the concatOutputs abstraction.

I'm wondering if it may just make more sense to have a cache input object, which represents an individual compileLoad entry?

This way we can only expose the `compileLoad` method, and then have an option to `build` and `buildSFX` to provide a compileLoad cache, so we don't need to expose `concatOutputs` as an API at all.

Something like - 

```javascript
  builder.compileCache = { loadname: outputCacheObj, ... }
```

Where `compileCache` is automaticaly populated / used during builds anyway by `compileLoad` and cleared by `builder.reset`.

Then `builder.compileCache` could be populated manually as a first API, and we can seek to improve this over time. Suggestions for something better are welcome too.

guybedford:
Actually the compileCache would probably be populated and used by compileOutputs.

ohjames:
An abstraction over a set of concatenated modules would be useful, I'd provided a thin layer in my build system around the existing builder to do exactly that and I think it probably belongs in `builder` itself. I stuck the `output` as an extra property on the `load` to maintain the association between source and output, that's pretty important to be able to do and yet the `compileLoad`/`concatOutputs` API doesn't do this which is a serious deficit.

So maybe this would work better as a layer on top of builder?

```javascript
var builder = new Builder(...)
var compileCache = new CompileCache(builder, opts)

compileCache.addLoads([ load1, load2 ])

// uses builder.compileLoad for each added load and then returns result
// of builder.concatOutputs
compileCache.compile().then(outputs => {
  // ...
  writeOutputs(...)
})
```

Even though the `concatOutputs` API doesn't make the best exported API for a caching builder, it still cleans up the existing `compileOutputs` code and it's an added bonus if it can be used simply by a caching layer on top.

ohjames:
You'd also want `removeLoads()` in addition to `reset()` for when files are deleted.

ohjames:
It's probably better to just add caching as a feature to the `builder` itself though as you initially suggested, would be a bit simpler to deal with and I guess in the majority of cases you either want the cache or don't care if it is maintained. Should be easy for me to implement any of these ideas.

guybedford:
So how about something like - 

```javascript
builder.setCache({
  get: function(name) {
    return output || undefined || Promise(output || undefined);
  },
  set: function(name, output) {
    // called along with compileLoad internally
    return undefined || Promise(undefined);
  }
});

builder.clearCache(path / wildcard path / undefined);
```

guybedford:
Uggh, still not happy...

ohjames:
If you want to be flexible with the caching then maybe you could just pass the cache as an argument to `compileLoad`, then have another method `compileCache(cache)` that can take such a cache.

ohjames:
```javascript
var cache = opts.cache = new LoadCache
builder.compileLoad(load1, opts)
builder.compileLoad(load2, opts)
builder.compileCache(cache)
```

ohjames:
Or maybe that should be `concatCache` instead of `compileCache`.

ohjames:
I updated my sample code... I kinda like it because then removing an object is as simple as `delete cache[load.name]`.

guybedford:
That seems ok, but I really don't want to have a concat API. I do think we need something that covers the whole use case of caching including invalidation, which does require a bit more thought.

ohjames:
I guess for me I'd like to see a minimal possible API that is flexible enough to support build systems. With browserify you see it taking on more and more of the role of a build system which seems to partly go against their stated goal of modularity.

Without a `concat` stop you could just have `compileLoads` which takes many loads + the same cache as before, the cache and the load could store the revision of the input/output so that this function could determine when to invalidate (or the cache could just store a checksum of the input or something). You could also potentially cache the output in the load itself, maybe that's ugly though.

ohjames:
At least if the load contained the cached output then removing the load would automatically remove it's cache entry, updating the load could clear the cached output requiring no checksum etc. but of course the code is more tightly coupled then.

guybedford:
There are actually two caches though - the trace cache, and the compile cache. Both are keyed by URL.

Repopulating the trace cache in the loader is not currently possible, so we probably initially want to skip this and just let the loader touch the filesystem again. Although this should be possible in future.

The compile cache is then the easy first step - but it may not be that much of a saving in comparison to what you're doing right now.

ohjames:
This compile cache already exists or are you talking about a new compile cache to be used by this new API?

If so I kinda like the `compileLoads` idea, maybe instead of just taking an `Object` as a cache it could take a `CompileCache` object to be a bit more flexible?

Or maybe something entirely different?

guybedford:
For now to be able to move forward with this, I'm happy to expose the compileLoad API, but not the concatOutputs one. Lets keep concatOutputs as semi-private so you can still use it in your work, then we can at least merge this and come back to the bigger questions separately?

ohjames:
Cool thanks for being so flexible on this, but I figure I don't like `concatOutputs` either now, so even making it semi-public would kinda suck.

Instead I just want to pass an optional cache parameter to `builder.buildTree` (and `compile.compileOutputs` which still remains private), this will also make it much simpler for build systems to interact with.

guybedford:
Sure, that sounds reasonable.

ohjames:
Cool just committed the (fully backwards compatible) caching API, let me know what you think. If it's acceptable I can add some mocha tests so it can be pulled.

lib/builder.js
```diff
@@ -240,7 +240,10 @@ function addExtraOutputs(output, tree, opts) {
   });
 }
 
-Builder.prototype.buildTree = function(tree, outFile, opts) {
+/**
+ * @param {Object} [cache] - Optional object used to cache module outputs.
+ */
+Builder.prototype.buildTree = function(tree, outFile, opts, cache) {
```
guybedford: Is it worth considering `opts.cache` here?

lib/compile.js
```diff
@@ -36,29 +36,55 @@ function compileLoad(loader, load, opts, compilers) {
...
 
+function retrieveCachedOutput(cache, load) {
+  if (! cache)
+    return;
+
+  var entry = cache[load.name];
+  if (! entry)
+    return;
+
+  return load.source === entry.source && entry.output;
```
guybedford: `entry.output` isn't set anywhere?

Perhaps we should do a hash of the input source for the cache object instead of storing the full source in the cache.

guybedford:
Nice, this seems like the right sort of approach.

lib/compile.js
ohjames: Oops ;) Fixed it, will push in 60 secs.

lib/builder.js
ohjames: Yeah I considered putting it there (as you can see from my earlier comments) but it felt more like "state" than "option", although it is kinda also "an option to use the cache" so can see it working either way.

ohjames:
Okay, fixed actually setting the `output` and pushed, so remaining:

1. Store hash of output instead of output itself.
2. Use opts.cache instead of extra argument (possibly).
3. Write regressions.

Then done! Let me know what you want for (2) and I'll finish it off tomorrow.

lib/builder.js
guybedford: That makes sense.

lib/compile.js
```diff
@@ -70,11 +97,20 @@ function compileOutputs(loader, tree, opts, sfxEntryPoints) {
...
+    if (output)
+      return output;
+
+    return compileLoad(loader, load, opts, compilers).then(function(output) {
+      updateCacheEntry(cache, load, output);
+      return output;
+    })
   }))
+  .then(function(loadOutputs) {
+    outputs = loadOutputs.filter(function(output) {
```
guybedford: Why do you need this filter?

lib/compile.js
```diff
@@ -36,29 +36,56 @@ function compileLoad(loader, load, opts, compilers) {
...
       throw new TypeError('"' + name + '" was defined via a bundle, so can only be used for subtraction or union operations.');
 
-    if (sfxEntryPoints && load.metadata.loader && load.metadata.build === false)
-      outputs.push('System.registerDynamic("' + load.name + '", [], false, function() { console.log("SystemJS Builder - Plugin for ' + load.name + ' does not support sfx builds"); });\n');
-
-    if (load.metadata.build === false)
-      return;
+    if (load.metadata.build === false) {
+      if (sfxEntryPoints && load.metadata.loader) {
+        return 'System.registerDynamic("' + load.name + '", [], false, function() { console.log("SystemJS Builder - Plugin for ' + load.name + ' does not support sfx builds"); });\n';
```
guybedford: This is not the same as the above.

ohjames: If you follow the promise chain you'll see it's equivalent and, IMO, a little neater.

lib/compile.js
ohjames: Due to line 86.

ohjames: Happy to revert back to pushing to `outputs` if you prefer it though, would be a little bit more optimal just feels less "functional", probably a matter of taste, I'm big into Haskell/lodash which possibly affects my js pragmatism.

guybedford:
Looks great! Happy to merge when its ready.

ohjames:
Great! It just needs a few mocha tests, will let you know when I've added the commit for that. Then not only will jspm + incremental rebuilds be a reality, they'll be the fastest incremental rebuilds around. That should silence the critics :) Will be blogging about it as soon as we're done.

guybedford:
Is this waiting review?

Just about to start some new work, don't want to mess with the diffs again!

ohjames:
Thanks for the heads up! I just wanted to add some new tests to prove the cache is being used.

I'd planned to add those later this evening, up to you if you want to merge this code before I've written the new tests. I'm happy to rebase against the upstream release if you wanna wait for them, also cool with opening a new pull request with the tests.

guybedford:
Thanks for being flexible. 

The work I'm attempting actually needs the compileOutputs function to be separated out into compileLoads and concatOutputs after all... I'm so sorry to be a pain!!

You don't still have a record of that work do you, as it would actually save me time now!?

ohjames:
Hehe, yeah I have a copy of it. I've just rebased it, I'll try and merge them both together.

ohjames:
You still need `compileLoad` and `concatOutputs` exported via `builder`?

guybedford:
Thanks - no need to worry now, I think I'm deep enough into it at this point. Will let you know when I'm done meddling.

ohjames:
Then I will play Splatoon.

guybedford:
@ohjames hope you are winning over there...

Don't hate me, but I've just merged a refactor in https://github.com/systemjs/builder/commit/7a71fdb39a98adc219fc7daa5d7a14ce190e3aa1.

I think it may make things simpler? I don't plan on meddling anymore for now.

ohjames:
Simpler than it already being finished? Hehe, doubt it :palm_tree:

ohjames:
Okay I'll give this one more try today :baby_chick:

ohjames:
Cool that only took 10 minutes, just need to add the tests now.

ohjames:
BTW did you know with mocha you can return a promise directly instead of having to bother with the `done` argument?

These tests don't seem to be taking advantage of that, all the `.catch(done)` etc. could be remove.

ohjames:
Okay test is committed also, should be ready for merge.

ohjames:
Made it use a hash of the content rather than the content and improved the tests a bit.

ohjames:
Okay finally stopped messing with it now.

README.md
```diff
@@ -253,6 +253,29 @@ Promise.all([builder.trace('app/first.js'), builder.trace('app/second.js')])
 });
 ```
 
+### Using the compiler cache
```
guybedford: With new features like this I prefer not to document them in code over the readme, so that we aren't stuck with the API just yet.

lib/builder.js
```diff
@@ -262,7 +265,10 @@ Builder.prototype.buildTree = function(tree, outFile, opts) {
   });
 };
 
-Builder.prototype.buildSFX = function(expression, outFile, opts) {
+/**
+ * @param {Object} [cache] - Optional cache for storing/reusing compiled modules.
+ */
+Builder.prototype.buildSFX = function(expression, outFile, opts, cache) {
```
guybedford: Can we also add cache to `Builder.prototype.build`.

guybedford:
This looks great! Think we're nearly there on the merge!

README.md
ohjames: Did you mean "to" rather than "not to" I'm not sure which way to make sense of that sentence.

guybedford: Right, yes exactly - not in the readme.

ohjames:
Cool, removed the docs from the README, added some docs to the code, added the cache to `build`, should be done?

guybedford:
Great, thanks for your hard work on this! If you do any benchmarks on the incremental building, please do share.

ohjames:
Cool I'll be writing a blog post about it very soon. If it's possible, could you release a beta version of systemjs-builder to npm containing this change? And a new beta version of `jspm` that includes that version of systemjs-builder?

That would make it much easier for others to take advantage of the work, if not it's no big deal, I can just have people install `systemjs-builder` from git and they'll just have duplicate versions of systemjs-builder, that version and the `jspm` dependency.

guybedford:
Sure, I've released this in 0.12.1, and it should be used in new installs of the beta now as well.

PR #195, opened 2015-06-12
pass load.name as module name for TypeScript

vladima:
pending items:
- [ ] - tests


guybedford:
Excellent, thanks. Which TypeScript version will this work against?

vladima:
Support for named modules will be in TypeScript 1.5 which is the upcoming release.

Issue #214, opened 2015-06-29
Overwriting exports?


ahmedshuhel:
Aren't we overwriting [this export ](https://github.com/systemjs/builder/blob/master/lib/builder.js#L11) with [this one](https://github.com/systemjs/builder/blob/master/lib/builder.js#L482) ? 

Shouldn't it be like:
```javascript
module.exports.getCanonicalName = getCanonicalName
module.exports.Builder = Builder;
```
Although, this may break a lot of other libs that depends on this. 
Could we add `getCannonicalName` func to `Builder's` prototype instead?

guybedford:
Yes that would be a mistake!

Happy for it to be a static though, so not part of the prototype - `Builder.getCanonicalName`.

guybedford:
Actually sorry, yes you're completely right - a prototype method would make the most sense as it needs the loader instance.

ahmedshuhel:
Should I make a PR for this? Right now, Is there anything other than [arithmetic.js](https://github.com/systemjs/builder/blob/master/lib/arithmetic.js#L103) depends on it?

guybedford:
@ahmedshuhel that would be amazing if you can.

ahmedshuhel:
Great! I am on it.

ahmedshuhel:
@guybedford Please review. 

Would it be better if we factor out [getCanonicalName](https://github.com/ahmedshuhel/builder/blob/57332989537f877f8eda90769327e5cd7ade4541/lib/builder.js#L40) from `builder` to a separate module? It may be useful in some cases to use a `static` function like this. 

 We may place it in `utils.js` as this houses all the static functions at the moment.

guybedford:
We're trying to avoid statics as it adds unnecessary complexity to the API.

guybedford:
Thanks, looking good.

ahmedshuhel:
@guybedford Inlined the func. Thank you.

guybedford:
@ahmedshuhel thanks for your help pushing this through.

PR #215, opened 2015-06-29
fix `getCanonicalName` exports overwriting #214

ahmedshuhel:


ahmedshuhel:
@guybedford Inlined as you have suggested. I rebased too but don't know why github showing your changes also. Please let me know if it's something I did.

guybedford:
@ahmedshuhel unfortunately this looks like bad timing with some of my work today, as I force pushed the release commit when I made a mistake. I've manually merged in https://github.com/systemjs/builder/commit/c0804dfabba282ef228b372893e06a537d25c63a.

guybedford:
@ahmedshuhel somehow it looks like the manual merge didn't catch the removal

PR #281, opened 2015-08-19
Re-throw uglify parse error

### LINKED ISSUE #280: uglifyJS throws bad formated error
Uglify Parse error makes debugging impossible. 
As mentioned here https://github.com/terinjokes/gulp-uglify/issues/76 the UglifyJS has a bad formatted JS_Parse_Error that needs to be catched and handled manually.

When using CLI's thats not a real option so the console output needs to be readable.

Psvensso:
Fixes #280


guybedford:
Thanks!

PR #285, opened 2015-08-21
extraneous comma in global expression builder

dennispg:
specifying more than one global in the meta config breaks the builder.


guybedford:
Thanks for catching this!

PR #294, opened 2015-08-27
Fix undefined `hasOwnProperty` in IE8

richardTowers:
IE < 9 does not define `hasOwnProperty` for host objects (see http://stackoverflow.com/a/8159434). This seems to have been fixed in systemjs core already (see https://github.com/systemjs/systemjs/commit/1ee77a#diff-5010b5e8eb34234db769e98607569c59L6).

I wasn't sure if there was some automatic way of updating the helper from `systemjs/systemjs` so I've only touched the line that's broken. Let me know if there's anything else I need to do.

### DIFF
--- a/templates/global-helpers.js
+++ b/templates/global-helpers.js
@@ -1,6 +1,6 @@
 (function(__global) {
   var loader = $__System;
-  var hasOwnProperty = __global.hasOwnProperty;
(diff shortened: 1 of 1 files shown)


guybedford:
Excellent thanks, must have missed this in the updates.

PR #309, opened 2015-09-10
Bundle source order in post order traversal order

Munter:
This PR fixes indeterministic source order in bundles, which means that bundle filename hashing will be consistent for unchanged sources.

Further more it order the sources in a way so dependencies for modules always comes before the module that requires it. In cases where parsing the bundle itself causes side effects by one module, which a module later in the load order depends on, bundling does not break these implicit side effect dependencies. This improves robustness for compatibility with non-module loader code.

Tests are not passing yet. I have one or two quirks that I still need to resolve

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -202,7 +202,7 @@ function getGraphEntryPoints(graph, entryPoints) {
   return entryPoints;
 }
 
-exports.postOrderTraverseTree = function postOrderTraverseTree(tree, entryPoints) {
+exports.getTreeModulesPostOrder = function getTreeModulesPostOrder(tree, entryPoints) {
   entryPoints = entryPoints || [];
 
   // Post order traversal sorted module list
@@ -211,14 +211,14 @@ exports.postOrderTraverseTree = function postOrderTraverseTree(tree, entryPoints
 
   // Seed graph with all relations
   Object.keys(tree).forEach(function (moduleName) {
-    var module = tree[moduleName];
+    var load = tree[moduleName];
 
-    if (!graph.adjList[module.name]) {
-      graph.addVertex(module.name);
+    if (!graph.adjList[moduleName]) {
+      graph.addVertex(moduleName);
     }
 
-    module.deps.forEach(function (depName) {
-      graph.addEdge(module.name, module.depMap[depName]);
+    load.deps.forEach(function (depName) {
+      graph.addEdge(moduleName, load.depMap[depName]);
     });
   });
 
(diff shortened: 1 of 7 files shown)


Munter:
I think the breaking tests are caused by my assumption that `compileOutputs` would always be supplied with `sfxEntryPoints`, which does not seem to be the case.

guybedford:
Do you still need `sfxEntryPoints` as an input if you're running the entrypoints detection on the graph anyway?

Munter:
I'm using `sfxEntryPoints` to determine as much ordering as possible. Think I got the error though. Forgot to add nodes to the graph in cases where the trace contains no dependencies

Munter:
Alright! Next step: Add tests for source ordering in bundles

lib/utils.js
```diff
@@ -163,4 +166,67 @@ function getAlias(loader, canonicalName) {
...
+
+  // Seed graph with all relations
+  Object.keys(tree).forEach(function (moduleName) {
+    var module = tree[moduleName];
+
+    if (!graph.adjList[module.name]) {
+      graph.addVertex(module.name);
+    }
+
+    module.deps.forEach(function (depName) {
```
guybedford: When we merge this with the conditional builds work, this can be replaced by `require('trace').getLoadDependencies(module)`.

lib/utils.js
```diff
@@ -163,4 +166,67 @@ function getAlias(loader, canonicalName) {
...
+exports.postOrderTraverseTree = function postOrderTraverseTree(tree, entryPoints) {
+  entryPoints = entryPoints || [];
+
+  // Post order traversal sorted module list
+  var postOrder = [];
+  var graph = new Graph(true);
+
+  // Seed graph with all relations
+  Object.keys(tree).forEach(function (moduleName) {
+    var module = tree[moduleName];
```
guybedford: I'd suggest calling this `load` as it is a load record.

guybedford: Perhaps this is confusing terminology though - it's just that the concepts align with the original spec concepts.

lib/utils.js
```diff
@@ -163,4 +166,67 @@ function getAlias(loader, canonicalName) {
...
+  }).sort().forEach(function (moduleName) {
+    if (entryPoints.indexOf(moduleName) === -1) {
+      entryPoints.push(moduleName);
+    }
+  });
+
+  return entryPoints;
+}
+
+exports.postOrderTraverseTree = function postOrderTraverseTree(tree, entryPoints) {
```
guybedford: Perhaps call this `getTreeModulesPostOrder` or something like that? Strictly we're not really traversing, but ordering?

lib/utils.js
```diff
@@ -163,4 +166,67 @@ function getAlias(loader, canonicalName) {
...
+    module.deps.forEach(function (depName) {
+      graph.addEdge(module.name, module.depMap[depName]);
+    });
+  });
+
+  // Post order traversal of graph, one per entryPoint
+  getGraphEntryPoints(graph, entryPoints).forEach(function (entryPoint) {
+    depthFirst(graph, entryPoint, {
+      leaveVertex: function (moduleName) {
+        // Avoidj duplicates
```
guybedford: Typo

lib/utils.js
```diff
@@ -163,4 +166,67 @@ function getAlias(loader, canonicalName) {
...
+          postOrder.push(moduleName);
+        }
+      }
+    });
+  });
+
+  // The graph some times includes things the tracer has already discarded. Filter those
+  postOrder = postOrder.filter(function (moduleName) {
+    return moduleName in tree;
+  });
```
guybedford: This check can be moved up above to when we push to `postOrder`?

Munter:
All changes in comments should be addressed now

guybedford:
:+1:

PR #312, opened 2015-09-15
Option to transpile cjs to es5 with traceur

hmalphettes:
Now that nodejs v4 is out nodejs developers are combining es6 with commonjs modules.
The option `cjs-to-es5` let traceur transpiles that code into es5.

This is useful for:
- minification as uglify-js only supports es5
- supporting other browsers than Chrome.

### DIFF
--- a/compilers/amd.js
+++ b/compilers/amd.js
@@ -250,7 +250,7 @@ exports.attach = function(loader) {
         // NB can remove after Traceur 0.0.77
         if (!load.source) load.source = ' ';
         var compiler = new traceur.Compiler({ script: true, sourceRoot: true });
-        load.metadata.parseTree = compiler.parse(load.source, load.address);
+        load.metadata.parseTree = compiler.parse(load.source, load.path);
         var depTransformer = new AMDDependenciesTransformer();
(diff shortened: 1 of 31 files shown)


guybedford:
Thanks for posting. SystemJS will be moving towards https://github.com/systemjs/systemjs/issues/579. 

Unfortunately a patch like this will fail when using transpilers. Also we need to make sure that build features always match SystemJS browser features, so can't create a difference here either.

hmalphettes:
@guybedford thanks for the review.
So what is the best way to proceed until https://github.com/systemjs/systemjs/issues/579 lands?

hmalphettes:
@guybedford "this will fail when using transpilers".
I am not sure I understand that and that is probably good news; maybe I am missing the obvious. When I run `jspm bundle` what should I also do to transpile commonjs-require with es6 into es5 code?

guybedford:
I mean that a lot of users use Babel for compilation instead of Traceur, while this approach only uses Traceur for the transpiling. For now I would suggest running this as a custom precompilation step, perhaps converting `app` to `app-compiled` or something like that.

hmalphettes:
Thanks for your patient explanation.

PR #316, opened 2015-09-19
Add missing saveForReset parameter

MeoMix:
saveForReset was not beign passed from executeConfigFile to builder.config. This was causing build.config to work improperly.


guybedford:
Thanks!

PR #341, opened 2015-10-05
Builder invalidation API

guybedford:
`builder.invalidate('path')` or `builder.invalidate('some/wildcard/*')`

Invalidation works in full module name space identical to the loader.

Invalidation is synchronous and returns the array of fully-normalized names invalidated - eg `['file:///path/to/project/x.js']`.

Inspired by the work of @jackfranklin in https://github.com/jackfranklin/jspm-dev-builder/blob/master/index.js!

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -44,6 +44,62 @@ function Builder(baseURL, cfg) {
     this.loadConfigSync(cfg, true, !!baseURL);
 }
 
+// invalidate the cache for a given module name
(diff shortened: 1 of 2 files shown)


guybedford:
A build workflow around this would be like:

```javascript
var builder = new Builder('.', 'config.js');

exports.build = function build() {
  builder.build(module, opts);
}

exports.invalidate = function invalidate(path) {
  builder.invalidate(path);
}
```

jackfranklin:
Awesome!

PR #362, opened 2015-10-20
Add fetch hook into tracer.

### LINKED ISSUE #330: Supporting source string as input
This would be the ability to provide a string as input into the builder that is JavaScript source code, i.e.:
```javascript
var script = "import foo from 'bar.js'; function useBar() { bar(); }";
var builder = new Builder('', 'system.conf.js');
builder.buildStatic(script);
```

The reason for this would be for build systems such as gulp, where you can chain multiple commands together. In my instance I am trying to build angular templates, attach them to the initial source file, then use system to build my dependency tree.

Thoughts?

dristic:
This is a first draft on how to hook static string data. This would allow you to build the string source of a file yourself, then pass it into the builder to resolve dependencies. Example:

```javascript
builder.build('foo.js', 'bar.js', {
  fetch: function (data) {
    if (data.name.indexOf('foo.js') !== -1) { return 'var bar = 2;'; }
  }
});
```

Fixes #330

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,28 @@ builder.config({
 });
 ```
 
+### Overriding Fetch
+
+The framework fetch function can be overridden in order to provide the source for a file manually. This is useful if you want to pre-process the source of a file before using the builder.
+
+```javascript
+var mySource = 'import * from foo; var foo = "bar";'; // get source as a string
+builder.bundle('foo.js', {
+  fetch: function (load, fetch) {
+    if (load.name.indexOf('foo.js') !== -1) {
+      return mySource;
+    } else {
+      // fall back to the normal fetch method
+      return fetch(load);
+    }
+  }
+});
+```
+
+The `load` variable describes the file that is trying to be loaded. This is called once for every file that is trying to be fetched, including dependencies.
+
+The `fetch` function should return a string.
+
 ### Bundle Arithmetic
 
 Both `builder.build` and `builder.buildStatic` support bundle arithmetic expressions. This allows for the easy construction of custom bundles.
--- a/lib/arithmetic.js
+++ b/lib/arithmetic.js
@@ -136,7 +136,7 @@ exports.traceExpression = function(builder, expression, traceOpts) {
           return getTreeModuleOperation(builder, op.operator)(curTree, op.moduleName);
         
         // tree . tree
-        return builder.tracer.traceModule(op.moduleName, traceOpts.traceAllConditionals, traceOpts.conditions, traceOpts.traceConditionsOnly)
+        return builder.tracer.traceModule(op.moduleName, traceOpts.traceAllConditionals, traceOpts.conditions, traceOpts.traceConditionsOnly, traceOpts.fetch)
         .then(function(nextTrace) {
           return getTreeOperation(op.operator)(curTree, nextTrace.tree);
         });
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -242,8 +242,7 @@ function processTraceOpts(options, defaults) {
     node: undefined,
     traceAllConditionals: true,
     conditions: {},
-    traceConditionsOnly: false,
-    fetch: undefined
+    traceConditionsOnly: false
   };
 
   extend(opts, defaults);
@@ -409,6 +408,20 @@ Builder.prototype.bundle = function(expressionOrTree, outFile, opts) {
     this.config(opts.config);
 
   var outputOpts = processOutputOpts(opts, { outFile: outFile });
+
+  // override the fetch function if given
+  if (opts) {
+    var loaderFetch = self.loader.fetch;
+    self.fetch = opts.fetch;
+    self.loader.fetch = function (load) {
+      var args = Array.prototype.slice.call(arguments, 0);
+      args.push(function () {
+        return loaderFetch.apply(self.loader, args);
+      });
+
+      return (self.fetch || loaderFetch).apply(self.loader, args);
+    };
+  }
 
   return Promise.resolve()
   .then(function() {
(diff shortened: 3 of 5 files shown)


test/test-build.js
```diff
@@ -39,6 +39,18 @@ function doTests(transpiler) {
...
 
+  test('Static string build', function () {
+    builder.reset();
+    builder.config({ transpiler: transpiler });
+    return builder.bundle('foo.js', {
+      fetch: function (data) {
+        if (data.name.indexOf('foo.js') !== -1) {
+          return fs.readFileSync('test/fixtures/test-tree/amd-1.js', 'utf8');
+        }
+      }
```
guybedford: Nice API! How about injecting this into the loader itself by implementing in the `lib/builder.js` reset() function - `var loaderFetch = builder.loader.fetch; builder.loader.fetch = function(load) { return (builder.fetch || loaderFetch).call(this, load); }` and then having this config option set `builder.fetch`?

dristic: Thanks! Do you think it is good practice to override the loader's functions like that? It might make it less predictable what the fetch function does in other areas of the framework, making it less modular.

What do you think about adding a fallthrough case as well?
```javascript
function (load) {
  var ret = builder.fetch(load);
  if (!ret) ret = loaderFetch(load);
  return ret;
}
```

This way it will fall through to using the default loader if you do not return anything from your `fetch` function.

guybedford: The suggestion above is based on a single extension of the loader that checks for a `builder.fetch` instance function on the builder not the loader itself. That is, setting `builder.fetch = null` completely reverses the fetch hook override. I'd prefer not to provide a fallthrough unless there is a good use case to justify it?

dristic: My only thought was being able to provide a complex expression or a glob, then provide the source as a string to a select few files. I could not be entirely understanding the system though.

I will look into the suggestion a bit more today and try to update this PR :smile:

dristic:
Okay I refactored it a bit to use the fetch override. One of the issues is that since we are completely overriding it, you can't pass in a string and have systemjs build the deps without the pass function. This kind of defeats the purpose so I added in a `pass()` that lets you fall back to the default implementation.

guybedford:
Thanks, this looks great. Agreed about the pass through concept - perhaps we could use the same pattern as we do for plugins then - https://github.com/systemjs/systemjs/blob/master/docs/creating-plugins.md#sample-css-plugin where the original `fetch` becomes the argument?

dristic:
Ah okay, I did not see this! I was modeling it after the `beforeEach(function (done) { done(); })` pattern from jasmine. I will update the PR and also add to the readme. :+1:

guybedford:
Great, thanks, just let me know when its ready as new commits don't send notifications.

dristic:
Sorry for the delay on this. Updated my code and the README as well!

lib/trace.js
```diff
@@ -35,7 +35,7 @@ Trace.prototype.traceModule = function(moduleName, traceAllConditionals, conditi
   })
   .then(function(canonicalConditionalEnv) {
     if (!conditionsOnly)
-      return self.getAllLoadRecords(moduleName, traceAllConditionals, canonicalConditionalEnv);
+      return self.getAllLoadRecords(moduleName, traceAllConditionals, canonicalConditionalEnv, undefined);
```
guybedford: This shouldn't be needed?

lib/arithmetic.js
```diff
@@ -136,7 +136,7 @@ exports.traceExpression = function(builder, expression, traceOpts) {
           return getTreeModuleOperation(builder, op.operator)(curTree, op.moduleName);
         
         // tree . tree
-        return builder.tracer.traceModule(op.moduleName, traceOpts.traceAllConditionals, traceOpts.conditions, traceOpts.traceConditionsOnly)
+        return builder.tracer.traceModule(op.moduleName, traceOpts.traceAllConditionals, traceOpts.conditions, traceOpts.traceConditionsOnly, traceOpts.fetch)
```
guybedford: This shouldn't be needed either?

lib/builder.js
```diff
@@ -409,6 +409,25 @@ Builder.prototype.bundle = function(expressionOrTree, outFile, opts) {
...
 
+  // override the fetch function if given
+  if (opts) {
+    var loaderFetch = self.loader.fetch;
+    self.fetch = opts.fetch;
+    self.loader.fetch = function (load) {
+      var args = Array.prototype.slice.call(arguments, 0);
+      args.push(loaderFetch.bind(self.loader));
+
+      var source = (self.fetch || loaderFetch).apply(self.loader, args);
```
guybedford: I think it's fine to use `(self.fetch || loaderFetch).call(self.loader, load, loaderFetch)` which cuts out the above two lines.

lib/builder.js
```diff
@@ -409,6 +409,25 @@ Builder.prototype.bundle = function(expressionOrTree, outFile, opts) {
...
+
+      if (typeof source === 'undefined') {
+        console.warn('Provided fetch function but did not return source for', load.name);
+        source = '';
+      }
```
guybedford: :+1:

guybedford:
Thanks for following up here!

README.md
```diff
@@ -187,6 +187,28 @@ builder.config({
 });
 ```
 
+### Overriding Fetch
```
guybedford: Can we move this down to the bottom of the docs, as it is an advanced feature?

dristic:
Oops, must have had some leftover changes from my previous method of doing this. I will update the branch with the fixes requested.

lib/builder.js
dristic: I swear I had a good reason for using apply :smile: Will fix

Issue #250, opened 2015-07-28
API document


guybedford:
It's time to create a `docs/api.md` file as the readme has reached its limits.

guybedford:
-[] Custom fetch hook in builder

guybedford:
Along with conditional loading, we need to link the conditional not resolved error to the doc page on how to set conditions comprehensively (https://github.com/systemjs/builder/issues/504).

ainscore:
Is this something that can be started on? I'd be interested in putting up a first draft.

ainscore:
Think it would also be good to document the module tree output from trace

guybedford:
Checking off any of these would be amazing.

PR #373, opened 2015-10-29
Add fetch hook into tracer

dristic:
This is to clean up the comments from #362 

- Moved override fetch to bottom of readme
- Cleaned up extra fetch override parameters
- Used call instead of apply inside bundle function

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,28 +187,6 @@ builder.config({
 });
 ```
 
-### Overriding Fetch
-
-The framework fetch function can be overridden in order to provide the source for a file manually. This is useful if you want to pre-process the source of a file before using the builder.
-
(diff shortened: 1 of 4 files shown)


dristic:
The build failure is due to `nvm install iojs`. I do not think it is related to this PR.

guybedford:
Thanks, I actually cleaned this up already in https://github.com/systemjs/builder/commit/42330dd0794db1901fd50d421cef74df5da2630c.

For the documentation part, I've created a link to this in https://github.com/systemjs/builder/issues/250, for addition in the API doc.

dristic:
Awesome, thanks! Maybe I can look at the API docs if I have some free time.

PR #374, opened 2015-10-29
Watch and automatically re-bundle pattern

onlywei:
Added a unit test to convince myself that this works the way I believed it to, and added an example pattern to the README for anyone else doing what I am doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ To make a bundle that is independent of the SystemJS loader entirely, we can mak
 builder.buildStatic('myModule.js', 'outfile.js', options);
 ```
 
-This bundle file can then be included with a `<script>` tag, and no other dependencies would need to be included in the page. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -283,6 +283,43 @@ Promise.all([builder.trace('app/first.js'), builder.trace('app/second.js')])
...
+    var filesToWatch = Object.keys(tree).map(function (filePath) {
+      // Remove the SystemJS plugin syntax from the file path.
+      return path.join(BASE_PATH, filePath.split('!')[0]);
+    });
+
+    builder.bundle(tree);
+    watcher = gulp.watch(filesToWatch, watch);
+  });
+}
+watch();
```
guybedford: Thanks for posting. Unfortunately this feature is not yet stable so not planned to be documented until then.

guybedford:
I've included a note to this in https://github.com/systemjs/builder/issues/250.

PR #375, opened 2015-10-29
Add more tests for cache invalidation.

onlywei:
I needed these tests to convince myself that
this API works the way I believed it to.

This is just the unit tests commit part of https://github.com/systemjs/builder/pull/374, since we're not ready to add documentation to the feature yet.


guybedford:
Thanks, I will review this when I can.

PR #377, opened 2015-10-30
Added baseURL and clone of global.System.map in builder.js / executeConfigFile

typhonrt:
Added baseURL and clone of global.System.map to configSystem which is necessary to support access to normalized map paths in further adding mapped paths via additional config loading during the build process. In executeConfigFile the global.System is temporarily rewritten to only expose the config method. This pull request adds `baseURL` and a clone of `global.System.map`.

This allows the ability to create additional mapped paths which access the current normalized paths for dependencies handled by JSPM / SystemJS. An example of an additional config file (`config-app-paths.js`) in my framework TyphonJS which uses this change is:

```
/**
 * Loads mapped paths for TyphonJS in the browser and via Gulp / Node.js allowing normalized
 * dependencies to be used in defining further mapped paths.
 */
var System = System || global.System;

// Debug for testing
console.log('Config-App-Paths - 0 - System.baseURL: ' +System.baseURL);
console.log('Config-App-Paths - 1 - System.map: ' +JSON.stringify(System.map));

var pathParseCore = System.map['typhon-core-parse'];
var pathParsePremium = System.map['typhon-premium-parse'];

System.config(
{
   map:
   {
      'appconfig': 'config/production-config.js',
      ...
      'typhonUser': pathParsePremium +'/src/control/user/TyphonUser.js',
      'typhonUserCollection': pathParseCore +'/src/collections/ParseUsers.js',
   }
});
```

The Gulp task to kick off SystemJS Builder looks like this:
```
   var builder = new jspm.Builder('./config.js');

   builder.loadConfig('./config/config-app-paths.js').then(function()
   {
      ...
   }
```

It should be noted that when running in the browser the full System instance is available. During the build process this is not available therefore adding a copy of baseURL and map is rather useful. 

Adding this change to Builder also resolves my feature request for uninstalling silent:
https://github.com/jspm/jspm-cli/issues/1220

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -251,7 +251,9 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   var configSystem = global.System = {
     config: function(cfg) {
       builder.config(cfg, ignoreBaseURL);
-    }
+    },
+    baseURL: curSystem.baseURL,
+    map: extend({}, curSystem.map)
   };
   // jshint evil:true
   new Function(source.toString()).call(global);


guybedford:
Thanks for posting this PR. I think if we're going to go this route, it may make more sense just to fully hijack the `System.config` function and provide the full loader.

typhonrt:
I'm all for having access to the full loader if there isn't any particular reason for limiting access via Builder as it's available in the browser after all. It would be great to get this in an update "soonish" as it provides a nice workflow for loading extra mapped paths which include normalized dependency paths. IE I'll be posting a Backbone-Parse demos repo over the weekend which will use it and include it in a tutorial if available. Thanks for taking a look!

guybedford:
If you're able to put together a PR on this I'll gladly review it. I can't guarantee I can personally work on it too soon though unfortunately.

typhonrt:
I'm closing this PR as #378 supersedes this one.

PR #378, opened 2015-10-30
Enabled full System Loader support in executeConfigFile

typhonrt:
Enabled full System Loader support in `executeConfigFile` in `builder.js`. This allows multiple System.config invocations to have full access to the System Loader API. The use case for this commit is to be able to load additional config files which add additional used defined mapped paths that incorporate existing normalized paths from loaded package dependencies.

As requested in https://github.com/systemjs/builder/pull/377 this pull request exposes the entire System Loader API to config loading in the Builder API. In particular the functions `loadConfig` & `loadConfigSync` invokes `executeConfigFile`.

Please see #377 for full use case description.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -246,16 +246,26 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   if (saveForReset)
     this.resetConfig = source;
 
-  var builder = this;
-  var curSystem = global.System;
-  var configSystem = global.System = {
-    config: function(cfg) {
-      builder.config(cfg, ignoreBaseURL);
+  // Save existing System.config function.
+  var systemConfigFunc = global.System.config;
+
+  // Assign a new temporary function which filters the config data; see `Builder.prototype.config`.
+  global.System.config = function(config) {
+    var cfg = {};
+    for (var p in config) {
+      if (ignoreBaseURL && p == 'baseURL' || p == 'bundles' || p == 'depCache')
+        continue;
+      cfg[p] = config[p];
     }
-  };
+    // Invoke existing loader config function.
+    systemConfigFunc(cfg); 
+  }
+
   // jshint evil:true
   new Function(source.toString()).call(global);
-  global.System = curSystem;
+
+  // Assign back to System.config the original saved function.
+  global.System.config = systemConfigFunc;
 }
 
 Builder.prototype.loadConfig = function(configFile, saveForReset, ignoreBaseURL) {


lib/builder.js
```diff
@@ -246,16 +246,26 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   if (saveForReset)
     this.resetConfig = source;
 
-  var builder = this;
-  var curSystem = global.System;
-  var configSystem = global.System = {
-    config: function(cfg) {
-      builder.config(cfg, ignoreBaseURL);
+  // Save existing System.config function.
+  var systemConfigFunc = global.System.config;
```
guybedford: We can't assume that it is the global system.

Try using rather - `var System = this.loader`, so that it will naturally be in the function scope below anyway, unless a user explicitly writes `window.System.config(...)`. We can swap out the global otherwise of course.

guybedford:
:+1:

This will also need to be updated in https://github.com/jspm/jspm-cli/blob/master/lib/config/loader.js#L70.

typhonrt:
The function below works for me. Could you advise if this is the correct solution:
```
function executeConfigFile(saveForReset, ignoreBaseURL, source) {
  if (saveForReset)
    this.resetConfig = source;

  var System = this.loader;

  // Save existing System.config function.
  var systemConfigFunc = System.config;

  // Assign a new temporary function which filters the config data; see `Builder.prototype.config`.
  System.config = function(config) {
    var cfg = {};
    for (var p in config) {
      if (ignoreBaseURL && p == 'baseURL' || p == 'bundles' || p == 'depCache')
        continue;
      cfg[p] = config[p];
    }
    // Invoke existing loader config function.
    systemConfigFunc(cfg); 
  };

  // jshint evil:true
  new Function(source.toString()).call(global);

  // Assign back to System.config the original saved function.
  System.config = systemConfigFunc;
}
```

I am not exactly sure about any associated change in jspm-cli as you mentioned re:
https://github.com/jspm/jspm-cli/blob/master/lib/config/loader.js#L70

Could you describe what needs to change here?

guybedford:
That looks great, happy to merge.

The code copied is the same piece of code in jspm, which should provide an identical compatibility for project consistency.

guybedford:
(I'm not saying you have to provide a PR for that, I'm just noting that it has to be done too)

typhonrt:
Excellent. PR #379 supersedes this one and I think we are good to go.

PR #379, opened 2015-10-30
Enabled full System Loader support in executeConfigFile

typhonrt:
Enabled full System Loader support in `executeConfigFile` in `builder.js`. This allows multiple System.config invocations to have full access to the System Loader API. The use case for this commit is to be able to load additional config files which add additional used defined mapped paths that incorporate existing normalized paths from loaded package dependencies.

As requested in #377 this pull request exposes the entire System Loader API to config loading in the Builder API. In particular the functions `loadConfig` & `loadConfigSync` invokes `executeConfigFile`.

Please see #377 for full use case description. This PR supersedes both #377 and #378

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -246,16 +246,28 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   if (saveForReset)
     this.resetConfig = source;
 
-  var builder = this;
-  var curSystem = global.System;
-  var configSystem = global.System = {
-    config: function(cfg) {
-      builder.config(cfg, ignoreBaseURL);
+  var System = this.loader;
+
+  // Save existing System.config function.
+  var systemConfigFunc = System.config;
+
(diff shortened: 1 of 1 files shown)


typhonrt:
Thanks Guy! Have good weekend... Any idea when the next JSPM NPM version bump might be? Also noticed that you can update some dependencies. I think Bluebird has a new major release (3.0.2).

guybedford:
Sure no problem! Not sure what is using Bluebird!? For future reference - you can push new commits to the branch or even force push (one of the few times force pushing is a good thing!) to update the PR without creating a new one.

typhonrt:
Good to know on updating a PR. 

Ah the Bluebird usage is in a dependency:
node_modules/jspm/node_modules/jspm-npm/node_modules/request/node_modules/har-validator requires bluebird@'^2.9.30' but will load

I just get warnings as I install 3.0.2 as a devDependency in my repos.

PR #388, opened 2015-11-04
Add use strict checking to cjs bundles.

### LINKED ISSUE #387: Bundling CJS module breaks strict mode
Suppose you have a CommonJS-style module called `app.js` which has a `"use strict";` pragma at the top to enable strict mode:

```js
"use strict";
var something = require("something");
// ... etc.
```

Bundling this module will insert additional lines prior to the `"use strict";` pragma:

```js
System.registerDynamic("app/app.js", ["something"], true, function(req, exports, module) {
  ;
  var global = this,
      __define = global.define;
  global.define = undefined;
  ;
  "use strict";
  var something = req("something");
  // ... etc.
  global.define = __define;
  return module.exports;
});
```

This breaks the pragma and disables strict mode, since the pragma must be the first expression in the function block.

Currently, I am working around this problem by inserting a build step prior to bundling which just wraps every JS file in an IIFE. I don't think I'd recommend that as the official fix, though. It might be simpler to just check whether the module source matches `/^['"]use strict['"]/` and, if so, insert an identical expression prior to the `global`/`define` boilerplate header.

Thoughts?

dristic:
This fixes #387 

The traceur compiler is quite a pain! The output is extremely verbose. As per the comments on the issue I just added a simple use strict check that prints out `"use strict";` on the first line if it passes the check.

This is just a first draft, there is probably some improvement I can make with feedback :smile:

### DIFF
--- a/compilers/cjs.js
+++ b/compilers/cjs.js
@@ -3,6 +3,7 @@ var url = require('url');
 var traceur = require('traceur');
 var ParseTreeTransformer = traceur.get('codegeneration/ParseTreeTransformer.js').ParseTreeTransformer;
 var Script = traceur.get('syntax/trees/ParseTrees.js').Script;
+var FunctionBody = traceur.get('syntax/trees/ParseTrees.js').FunctionBody;
 var parseStatements = traceur.get('codegeneration/PlaceholderParser.js').parseStatements;
 var parseExpression = traceur.get('codegeneration/PlaceholderParser.js').parseExpression;
 var STRING = traceur.get('syntax/TokenType.js').STRING;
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks for the attempt here. Since we're dealing with the parse tree, this should done properly through the parsing process.

The example code to follow for this would be https://github.com/google/traceur-compiler/blob/master/src/codegeneration/MakeStrictTransformer.js, especially the use of `hasUseStrict` and `createUseStrictDirective` both of which should form the basis of this code.

This would be along the lines of `traceur.get('semantics/util.js').hasUseStrict` and `traceur.get('codegeneration/ParseTreeFactory.js').createUseStrictDirective`.

If you're able to work on this that would be amazing.

dristic:
Ahh okay. Thanks for the feedback, I had no clue where to add this but this example looks much better than what I did!

I will take a look and try to get to it this week.

dristic:
After much research I am actually beginning to understand Traceur a lot more. I did not realize that it is actually lexing/token parsing the code out to transpile it. Google is crazy smart!

I added the code to check for strict mode and add the new strict mode directive. That being said you now end up with two `"use strict";` directives, so we need to remove the original one somehow. I am not sure of the traceur method of doing this so I figured I would post some code and then ask :smile:

guybedford:
Excellent! Glad it made some sense. It is mostly the monumental work of http://github.com/arv.

To remove the existing `use strict`, it may be worth inlining the implementation of `hasUseStrict` from https://github.com/google/traceur-compiler/blob/master/src/semantics/util.js#L32, and modifying it to rather slice the script list before and after to get the right output.

guybedford:
I hope you don't mind but I went ahead and manually merged this in https://github.com/systemjs/builder/commit/733eccf478f207a55ac7c2ed0b3b6ecffb6c7244 to get it in the current release.

dristic:
No worries! I will branch again off master and add the use strict removal. Glad this made it in!

guybedford:
Sure, I managed to include the duplication removal as well, but if you see anything odd do let me know.

dristic:
Ah okay. Just saw that. Thanks for adding it in! I will look for something else to add then :smile:

Issue #391, opened 2015-11-08
Log error in compileTree


LongYC:
There is a long chain of Promise in `compileTree` dealing with modules and plugins which also means there is a high chance of things could go wrong here but there is no log or any other information available  when it happens. I would rather have something like:

```
Building the single-file sfx bundle for jsx/main.js...
[ 'jsx/main.js:5:3: Unexpected token <',
  'jsx/main.js:5:4: Unexpected token div',
  'jsx/main.js:5:8: Semi-colon expected',
  'jsx/main.js:5:34: Unexpected token >',
  'jsx/main.js:7:1: Unexpected token )' ]
```

then nothing at all. I propose add a `catch` to Promise chain and log the error.

guybedford:
If

guybedford:
It looks like you are getting errors swallowed somehow in jspm?

If that is the case it may be a jspm bug not propagating error messages, although this should be handled by the rejection handler in https://github.com/jspm/jspm-cli/blob/master/lib/bundle.js#L177.

LongYC:
Sorry for the silly PR. Yeah, it seems to happen when I try to bundle JSX with `require` instead of ES2015 `import` and somehow the error does not get propagated, it got lost somewhere in jspm's `ui.log`.

PR #392, opened 2015-11-08
Log error in compileTree.

LongYC:
Issue #391

There is a long chain of Promise in `compileTree` dealing with modules and plugins which also means there is a high chance of things could go wrong here but there is no log or any other information available  when it happens.


guybedford:
Thanks for the PR. Promise APIs should return error as rejections and not automatically log them though - that is up to the consumer to manage.

LongYC:
Sorry for the rushed PR, that was really bad.

Issue #384, opened 2015-11-02
builder 0.14.10 regression 


typhonrt:
Greets Guy et al,

I'm still trying to track things down, but a complicated (to me) side effect / regression affects the changes implemented for #379 between `0.14.9` and `0.14.10`. I have so far confirmed I can apply the #379 patch to 0.14.9 and everything works as expected. Running against 0.14.10 the build completes fine, but there is a runtime error. I have verified that the code surrounding #379 seems to be working correctly in 0.14.10, so my only current assumption is that there is another side effect / regression. #379 was in builder.js and I have validated that the regression still exists when swapping builder.js (0.14.9) with #379 applied to 0.14.10. It seems the problem lies with other changes beyond builder.js as things go, so is complicated for me to debug though I am going to continue investigating. 

UPDATE / BELOW is related to #386:
`------`
The error that I'm seeing in the static build is that it completes and on load it doesn't seem to be loading correctly the user mapped paths which depend on a previous config load. IE see the code example in #377 that should work. I am now examining a folder compare between 0.14.9 and 0.14.10. 

It should be noted that everything works fine when in the browser via full source, so definitely a builder issue. In my framework code for a web app an import of custom mapped paths that should be loaded and executed simply isn't. In this case it is initialization code for Parse, so the error happens quickly on load. Due to hoisted ES6 imports and / or how SystemJS Builder works it's hard to debug.
`------`

There is however an actual regression here described in follow up comments. It turns out that using `loadConfig` for additional config loading when queueing multiple Builder / bundle operations causes a failure / clobbered configuration data. More is described below.

Guy, could you offer any insight into any conceivable changes between these versions? I'm pretty sure the regression may be beyond `builder.js`. I'm fishing right now... 

Thanks...

typhonrt:
OK.. I think I've narrowed down the trigger condition. It turns out that the problem I'm seeing is due to my Gulp task which queues multiple bundle / build operations. Something is being reset / order of operations different internally to SystemJS Builder in 0.14.10 versus 0.14.9.  

BTW I have compiled a NPM package, [typhon-core-gulptasks](https://www.npmjs.com/package/typhon-core-gulptasks) of common Gulp tasks for use with JSPM / SystemJS. It definitely helps to share workflows across many projects and is really easy to import into projects. 

[...]

PR #394, opened 2015-11-09
Two builders calling loadConfig in parallel

fictitious:
This adds the testcase for two builders calling loadConfig in parallel, to support use case like this:

    var builder1 = new Builder();
    var builder2 = new Builder();

    builder1.loadConfig('config1.js').then(function() { ... });
    builder2.loadConfig('config2.js').then(function() { ... });

The test fails, and the second commit attempts to fix it.

See also #384 that might be related.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -264,7 +264,7 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   };
 
   // jshint evil:true
-  new Function(source.toString()).call(global);
+  eval(source.toString());
 
   // Assign back to System.config the original saved function.
   System.config = systemConfigFunc;
--- a/test/test-load-config.js
+++ b/test/test-load-config.js
@@ -1,0 +1,44 @@
+var Builder = require('../index');
+var fs = require('fs');
+var Promise = require('rsvp').Promise;
+
+
+suite('Test builder.loadConfig', function() {
+    
+  test('builder.loadConfig works', function(done) {
+    
+    var configFile = 'test/output/builderConfig.js';
+    var builder = new Builder();
+    fs.writeFileSync(configFile, 'System.config({map: {m: "./m.js"}});');
+    builder.loadConfig(configFile).then(function() {
+      
+      assert.equal(builder.loader.map['m'], './m.js', 'loader map was loaded from config');
+      
+    }).then(done, done);
+    
+  });
+  
+  test('builder.loadConfig does not affect other builders', function(done) {
+  
+    var configFile1 = 'test/output/builder1Config.js';
+    var configFile2 = 'test/output/builder2Config.js';
+    fs.writeFileSync(configFile1, 'System.config({map: {m1: "./m1.js"}});');
+    fs.writeFileSync(configFile2, 'System.config({map: {m2: "./m2.js"}});');
+    
+    var builder1 = new Builder();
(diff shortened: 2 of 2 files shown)


lib/builder.js
```diff
@@ -264,7 +264,7 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   };
 
   // jshint evil:true
-  new Function(source.toString()).call(global);
+  eval(source.toString());
```
guybedford: `new Function` is necessary to avoid global leaks.

guybedford:
Thanks this is a great help! I'll happily merge and resolve this.

lib/builder.js
fictitious: I see. Somehow the code must be executed in a way so that

     something = 'some value'; 

did not pollute global scope, but

    System.config(...);

referred to the local System declared in executeConfigFile.

I don't see any easy way to achieve both. Since there is transpiler available, one way to do it would be to call javascript parser, walk the AST and patch calls to System.config() to refer to the local System (which needs to be made accessible somehow), then generate source and call new Function().call() on it. But that seems like an overkill.

I'm happy to have this as wontfix, actually - I know how to use this API properly now, and parallel builds do not matter that much for me.

tyscorp: One possibility is to shim `System` in the `new Function` call. Eg:

```js
new Function(`
    var System = { confs: [], config: function (conf) { this.confs.push(conf); } };
    ${source.toString()}
    return System;
`).call(global);
```
The System shim would need be to be fleshed out much more and it would break if the config file had a `return` in it...

fictitious: Well telling people 'your config file must not have `return` at top level' would be quite reasonable IMHO. After all it's supposed to be included as script in web pages:

    <script src="config.js"></script>

is the way to use single `config.js` for both development and production workflow.

guybedford: Ahh right of course this was the underlying issue. Ok, try using `eval('(function() {' + source.toString() + '})()')` here then to avoid global variable assignments, but allow local variable access to `System`?

fictitious: I did some tests, I'm seeing that for both `eval(...)` and new `Function(...)` , assignment to undeclared variable causes creation of that variable in global scope. The only sure way to prevent that is to `use strict`, that is

`eval('(function() {"use strict"; ' + source.toString() + '\n})()')`

It would not prevent from assigning to global variables that already exist in global scope, however - I don't think there is any protection from that.

Now I'm wondering how many people have config files that will break in strict mode.

Thanks for the hint, I will update the PR with `eval function using strict`.

guybedford: How about just using `require('vm').runInContext(source, { System: ... })`?

fictitious: Yes this looks like the best one.

I think it would make sense to add at least console to the context. Will update the PR ...

guybedford: :+1:

guybedford:
I went ahead and merged so we can get this fix in the current patch release. The eval while not technically correct actually isn't that worrying to be honest.

fictitious:
Thanks! That's probably safer to do right before the release. 

I submitted a new PR with require('vm').executeInNewContext, as discussed - it might break people's config files, not sure if enforcing config file purity is worth it.

PR #400, opened 2015-11-16
fixed isLoadFresh accessed non existant loads variable

jakobwesthoff:
In certain cases (plugins) `isLoadFresh` tries to access the `loads` variable, which simply isn't present in the given context.

This PR fixes this issue.


guybedford:
Thanks, was just working on this!

PR #401, opened 2015-11-16
executeConfigFile: use require('vm').executeInNewContext

fictitious:
as discussed in   
https://github.com/systemjs/builder/pull/394#discussion_r44321335

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -275,8 +275,13 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
     systemConfigFunc(cfg); 
   };
 
-  // jshint evil:true
-  eval(source.toString());
+  // make everything in global available to config file code
+  // only substitute local copy of System
+  // and prevent that code from adding anything to the real global 
+  var context = Object.create(global);
+  context.System = System;
+  context.global = context.GLOBAL = context.root = context;
+  require('vm').runInNewContext(source.toString(), context);
 
   // Assign back to System.config the original saved function.
   System.config = systemConfigFunc;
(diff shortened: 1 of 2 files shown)


lib/builder.js
```diff
@@ -275,8 +275,7 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
     systemConfigFunc(cfg); 
   };
 
-  // jshint evil:true
-  eval(source.toString());
+  require('vm').runInNewContext(source.toString(), {System: System, console: console});
```
guybedford: I had a thought here. How about doing:

```javascript
var g = Object.create(global);
g.System = System;
```

That way we provide access to all globals, but deny writes.

guybedford:
Thanks for the follow-up!

fictitious:
Turns out it's slightly more convoluted. This is what I ended up with after doing some more tests.

guybedford:
Just to confirm I'm following correctly - `Object.create(global)` as the context fails for the `global.x = 'asfd'` case?

fictitious:
Yes - after

    var p = Object.create(global)

the following returns true:

    p.global === global

also, in node, there are also p.GLOBAL and p.root aliases referring to the same global. The code makes all these aliases refer to new context instead.

guybedford:
@fictitious since we know the global names in Node are `root`, `global`, `GLOBAL`, I'd prefer to just do this manually to avoid complete traversal of the global.

fictitious:
ok - assuming the risk of it ever changing in node is small enough

guybedford:
:+1:

PR #408, opened 2015-11-23
Invalidate throws if the cache is empty

jackfranklin:
@oliverjash and I spent ages debugging an issue only to realise it was our mistake and that the cache wasn't populated.

This PR changes the behaviour of invalidate to throw if the trace cache is empty. I understand that throwing might be controversial, and I'd happily consider swapping for `console.log`.

@guybedford what do you think? I think this could help with debugging.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -51,6 +51,10 @@ function Builder(baseURL, cfg) {
 // invalidate the cache for a given module name
 // accepts wildcards (*) which are taken to be deep-matching
 Builder.prototype.invalidate = function(invalidationModuleName) {
+
+  if (Object.keys(this.cache.trace || {}).length === 0)
+    throw new Error('Invalidate was called but the trace cache is empty');
+
   var loader = this.loader;
 
   var invalidated = [];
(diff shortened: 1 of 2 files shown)


guybedford:
Invalidate returns the array of modules that were invalidated - I think that should be enough information to debug if that list is empty?

jackfranklin:
That doesn't let you distinguish between the two scenarios:

1. You had the right module name, but the cache is empty
2. The cache is populated, but the module name isn't correct

This PR immediately spots and alerts you to (1), hopefully saving some debugging time. The more I think about it the more I'd swap the error for a console log mind.

guybedford:
Clearing a cache is a safe guaranteeing operation. That is I can run it aggressively to know that the cache has been cleaned. Throwing an error in some cases removes the ease of that use case. Imagine if clearing your trash can on a mac gave an error when there was no trash for example. Also the only time we output log messages is for deprecation warnings in this project.

If you're looking at avoiding your original mistake, it may help to understand what the original issue was that you still thought you needed to be clearing the cache properly?

jackfranklin:
@guybedford :+1: fair points, particularly on being able to invalidate regardless of the state of the cache.

I can't remember the exact issue @oliverjash and I had, but I remember commenting saying that if SystemJS had warned us the cache was empty it would have been useful.

PR #414, opened 2015-11-29
Take into account package-level default extension in getCanonicalName

### LINKED ISSUE #413: Builder drops file extension when js module is imported from jsx module
SystemJS Build Tool drops file extension when js module is imported from jsx module. I will illustrate the problem in a simple example.

There is a package `test-app` that sits in `lib` folder and consists of two modules `main.jsx` and `module.js`. Module `module.js` is imported into `main.jsx` module. In development mode (files are loaded separately to browser) SystemJS loads both modules as expected with correct extensions.

During build of self-executing bundle SystemJS builder is looking for `lib/module` file instead of `lib/module.js`:
```
err  Error on fetch for test-app/module at file:///c:/systemjs-builder-issue-import-js-from-jsx/lib/module
        Loading lib/main.jsx
        Error: ENOENT, open 'c:\systemjs-builder-issue-import-js-from-jsx\lib\module'
         at Error (native)
```
It is only reproducible when both `defaultExtension: false` option is set **and**  path to package is specified in `paths` option.
```javascript
packages: {
  "test-app": {
    main: "main.jsx",
    format: "esm",
    defaultExtension: false
  }
},

paths: {
  "test-app*": "lib*",
  "github:*": "jspm_packages/github/*",
  "npm:*": "jspm_packages/npm/*"
},
```
Particularly, when renaming `lib` folder to `test-app` and removing its entry from `paths`, build passes successfully. 

Pay attention that **SystemJS** and **SystemJS Build Tool** both work with the same configuration, but first handles it correctly and second fails.

See full example here: [maxlk/systemjs-builder-issue-import-js-from-jsx](https://github.com/maxlk/systemjs-builder-issue-import-js-from-jsx)

fictitious:
This is an attempt to fix #413

Not sure if it does not break other people setups though - loader.paths is populated from many places I think, the test covers config.paths only.

### DIFF
--- a/test/test-default-extension-path.js
+++ b/test/test-default-extension-path.js
@@ -1,0 +1,26 @@
+var Builder = require('../index');
+
+var builder = new Builder('test/fixtures/default-extension-path');
+
+suite('Test path to package with defaultExtension', function(err) {
+  test('path to package with defaultExtension', function() {
+      builder.config({
+          defaultJSExtensions: true,
+          paths: {
+            "test-app*": "lib*"
+          },
+          packages: {
+            "test-app": {
+              "main": "main.js",
+              "format": "esm",
+              "defaultExtension": false
+            }
+          }
+      });
+      
+      return builder.bundle('test-app').then(function(out) {
+          console.dir(out.entryPoints);
+          assert(out.entryPoints.length == 1 && out.entryPoints[0] == 'test-app/main.js');
+      });
+  });
+});
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -127,11 +127,20 @@ function canonicalizeCondition(loader, conditionModule) {
 function getCanonicalNamePlain(loader, normalized, isPlugin) {
   // now just reverse apply paths rules to get canonical name
   var pathMatch;
+  var defaultExtension = loader.defaultJSExtensions ? 'js' : false;
 
-  // if we are in a package, remove the basePath
+  // if we are in a package, remove the basePath and use its defaultExtension
   var pkgName = getPackage(loader.packages, normalized);
   if (pkgName) {
     var pkg = loader.packages[pkgName];
+    // use defaultExtension from the package
+    if (pkg) {
+      if (pkg.defaultExtension === false) {
+        defaultExtension = false;
+      } else if (pkg.defaultExtension) {
+        defaultExtension = pkg.defaultExtension;
+      }
+    }
 
     // sanitize basePath
     var basePath = pkg.basePath && pkg.basePath != '.' ? pkg.basePath : '';
@@ -151,7 +160,7 @@ function getCanonicalNamePlain(loader, normalized, isPlugin) {
     if (loader.paths[p].indexOf('*') != -1)
       continue;
 
-    var curPath = normalizePath(loader, p, isPlugin);
+    var curPath = normalizePath(loader, p, isPlugin, defaultExtension);
 
     if (normalized === curPath) {
       // always stop on first exact match
@@ -169,7 +178,7 @@ function getCanonicalNamePlain(loader, normalized, isPlugin) {
         continue;
 
       // normalize the output path
-      var curPath = normalizePath(loader, p, isPlugin);
+      var curPath = normalizePath(loader, p, isPlugin, defaultExtension);
 
       // do reverse match
       var wIndex = curPath.indexOf('*');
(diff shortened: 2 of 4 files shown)


test/test-default-extension-path.js
```diff
@@ -0,0 +1,26 @@
+var Builder = require('../index');
+
+var builder = new Builder('test/fixtures/default-extension-path');
+
+suite('Test path to package with defaultExtension', function(err) {
+  test('path to package with defaultExtension', function() {
+      builder.config({
+          defaultJSExtensions: true,
+          paths: {
+            "test-app*": "lib*"
```
guybedford: This should really be map config.

guybedford:
I think the issue here is that the paths config should really not trigger for empty wildcards... I believe replacing with map config should avoid the issue.

fictitious:
Yes with map it works - there is no matching entry in loader.paths, so getCanonicalName just removes baseURL from the front and does not change file extension. Testcase with map is on this branch https://github.com/fictitious/builder/commit/c848d65c0d01108ffed7e09e12692883796bc0d6 , it passes.

I think I understand why it does not work with paths. getCanonicalName is called with normalized,  absolute module url of the form `file:///<baseURL>/lib/main.js`.

It finds matching entry in loader.paths which has the form

      {"test-app*" : "lib*"}

It then calls normalizePath with the key, `test-app*` . The result is in a variable named curPath, it value is

      file:///<baseURL>/lib*.js

I think this is what causing the bug, because it's different from what loader.normalize() would have returned for that key - it has `.js` extension even though defaultExtension is set to false for this package (or path).

The next step is to find common suffix and prefix between `normalized` (the argument for getCanonicalName) and `curPath` , strip that suffix and prefix from `normalized`, and substitute what remains in place of wildcard in loader.paths key. 

In other words, because `curPath` has `.js` extension, it is getting stripped from normalized `main.js`, and the result of getCanonicalName does not match what's there in the file system.

With that limited understanding, I think the easiest way to make this case work is to replicate what loader.normalize() does in normalizePath() in utils.js - that's what my patch tries to do, and that's what comment in front of `getCanonicalName` seems to be saying

    // just reverse-applies paths and defulatJSExtension to determine the canonical

fictitious:
However all this is way too complicated, if you decide to change the rules for wildcards and simplify that stuff I'm all for it.

guybedford:
`normalizePath` is path normalization, which isn't supposed to apply package normalization.

So it's just determining that `file:///<baseURL>/lib*.js` is the normalization target for that path.

Ideally we would always say that the path target is without the `.js` extension, and that will be the case when defaultJSExtensions is deprecated as it is a big contributor to the normalization complexity currently.

I'm wondering if it might be possible just to do this already to entirely remove `defaultJSExtensions` but only in the wildcard cases so the single path equality still holds. Perhaps a flag to distinguish between wildcard and non wildcard would cover it.

guybedford:
That would effectively be to rename `isPlugin` to `skipDefaultJSExtensions` or similar, and set it to true for the wildcard case.

fictitious:
> normalizePath is path normalization, which isn't supposed to apply package normalization.

ah I see. I noticed the code that un-applies some package-level settings in getCanonicalName before normalizePath is called (it seems to be removing package basePath), I honestly did not know if package-level defaultExtension also needed to be handled there or not.

I see now that the reverse case - with `defaultJSExtensions: false`, and package-level `defaultExtension: js` just works and does not need any changes. 

So the problem will go away the moment defaultJSExtensions goes away. Disabling defaultJSExtensions for wildcard path match is one way to start moving in that direction. I tried to update the PR as you suggested, but with that change one test started failing - test/canonicals.js:30:12 (as expected - that test has wildcard paths, `defaultJSExtensions: true` and checks for jade plugin having `.js` in canonical name). I left it at this branch for now: https://github.com/fictitious/builder/commit/74ea0f39451e034069600454bb4bd435e79d4252

Aaand, I'm apparently pretty good at stumbling on edge cases - if I have the directory named lib.js, it fails with a different error.

fictitious:
Sorry just noticed that failing test is not passing isPlugin parameter to getCanonicalName, so the actual use case it test for is broken anyway. Perhaps all that needed is just updating that test?

guybedford:
@fictitious I think it would be valid to alter the test output in `https://github.com/systemjs/builder/blob/master/test/canonicals.js#L30` to `file.jade!jade.js`, but this would alter the naming of a lot of bundles so perhaps best not to make the change at this point after all and just let the deprecation happen?

fictitious:
Sure - when defaultJSExtensions is on the deprecation path, the fix should have no impact at all on other things, ideally. Closing.

PR #424, opened 2015-12-08
Fix empty sourceRoot for Typescript source-maps

justindujardin:
This PR addresses systemjs/systemjs#177 by fixing bad source file paths in output Typescript source-maps.  I'm not sure if this is the place to apply the fix, please provide feedback.

The issue is that the Typescript compiler is returning source maps with the paths stripped off, e.g. "source/api/point.ts" would have a sourcemap with no "sourceRoot" and "point.ts" as the file name.

### DIFF
--- a/compilers/esm.js
+++ b/compilers/esm.js
@@ -1,4 +1,5 @@
 var traceur = require('traceur');
+var path = require('path');
 
 var ParseTreeTransformer = traceur.get('codegeneration/ParseTreeTransformer.js').ParseTreeTransformer;
 var ModuleSpecifier = traceur.get('syntax/trees/ParseTrees.js').ModuleSpecifier;
@@ -165,7 +166,15 @@ exports.compile = function(load, opts, loader) {
       };
       
       var transpiled = transpiler.transpileModule(source, transpileOptions);
-      
+
+      // Typescript compiler source map has no "sourceRoot" on entries when
+      // transpiled in this way. Apply the sourceRoot to sourcemap manually.
+      if (transpiled.sourceMapText) {
+        var sourceMapObject = JSON.parse(transpiled.sourceMapText);
+        sourceMapObject.sourceRoot = path.dirname(load.path);
+        transpiled.sourceMapText = JSON.stringify(sourceMapObject);
+      }
+
       return Promise.resolve({
         source: transpiled.outputText,
         sourceMap: transpiled.sourceMapText


guybedford:
Thanks for tracking this down! I think we should fix this by supporting source maps without a source root rather in the resolution at https://github.com/systemjs/builder/blob/master/lib/sourcemaps.js#L48 to make the implementation more flexible.

justindujardin:
@guybedford Okay, I think I understand what you want.  I looked into it a bit and need clarification/direction.

1. Your link points to a piece of code inside a particular conditional for `sourceMapContents`.  I assume you meant this should belong in the Promise chain for processing source map files, and not specifically inside this conditional? If you did mean inside this conditional, please explain a bit about the `sourceMapContents` option.  It's tests/documentation don't make it very clear.

2. I'm having a hard time seeing how I can reconstruct the source map path inside of the sourcemaps promise chain.  It appears the compilers all return a promise with just two properties: `{source: '...', sourceMap: '...'}`. The reason I'm able to fix the path in the first place is because of the knowledge of the `load.path` value.  Should I update the compilers to pass this information along in a new property (`loadPath`?), or is there another way to get at the load object while processing source maps that I'm not seeing?

guybedford:
Thanks for the feedback.

1. I kind of blindly pointed to the only reference to `sourceRoot`, I think the point is that we need to better handle `sourceRoot` sanitization as part of the source maps generation process.

2. Every source map contains its original load.path in `sourceMap.name` in the source map itself as an absolute file path I believe.

I certainly don't expect you to fix all this, I'm just pointing where the issue is given this PR. Happy to create an issue to track sanitization of this area further.

justindujardin:
1. Great we're on the same page there.

2. I think what you point out is exactly what this is fixing.  The typescript compiler (for whatever reason) spits out a sourcemap that does not contain the full path to the input file.  Here's what a sourcemap for `pow-core/resources/audio.ts` looks like coming back from the compiler:

```json
{
	"version": 3,
	"file": "audio.js",
	"sourceRoot": "",
	"sources": ["audio.ts"],
	"names": ["AudioResource", "AudioResource.constructor", "AudioResource.supportedFormats", "AudioResource.fetch", "AudioResource.play", "AudioResource.pause", "AudioResource.volume", "AudioResource._loadAudioBuffer", "AudioResource._loadAudioElement"],
	"mappings": "..."
}
```

guybedford:
@justindujardin ah I see, this PR does sound like the better approach then to ensure the correct behaviour. Perhaps instead of setting `sourceRoot`, the `file` value can be made to be a full `load.path`?

justindujardin:
I gave a shot at setting the `file` value to `load.path` and it appears that `sourceRoot` is significant in some non-obvious way. The path looks correct in the modified source map for the individual files, but when they're concatenated the path appears to be lost:

*input* = `source/pow-core/resources/audio.ts`
*destination* = `lib/pow-core/pow-core.js.map`
*output sourceRoot* = `../../source/pow-core/resources/audio.ts`
*output file* = `../../source/audio.ts`

I don't see any usages (other than the one you originally pointed out) of `sourceRoot` in the builder project, so I must assume some other source map processing code is taking it into account.

guybedford:
@justindujardin I think that output is correct actually - `../../source/audio.ts` being the only way we can know how to trace back to the original source without any further information. This is done in https://github.com/systemjs/builder/blob/master/lib/sourcemaps.js#L87. What would you expect the output to be?

It may be worth having a `sourceRoot` option in the builder itself to alter this path, but by default it assumes the baseURL itself I believe.

chrismagri:
@guybedford when do you think this change will be released?

justindujardin:
@guybedford apologies--I forgot to follow up on this. 

The path `../../source/audio.ts` is not actually correct, so when it is resolved with the `basePath` in the builder, it points to a bad location.  The file exists in `../../pow-core/resources/audio.ts` where `pow-core/resources` is stripped off without the proper sourceRoot on the source map. I'm sorry I don't fully understand exactly why this is happening, but I have no more time to dig into it.

PR #431, opened 2015-12-15
Fixed accidental global var

papandreou:


guybedford:
Thanks! I've pushed out a 0.14.13.

PR #435, opened 2015-12-21
You're using bluebird, you might as well enjoy it a little.

benjamingr:


guybedford:
Thanks for the PR. I'd prefer to stick to the spec API so that it is always easy to change the promises implementation though.

PR #437, opened 2015-12-22
bug fix in register.js

catgofire:
tree.args is an Object and has no method "concat"... tree.args.args is the array we're looking for.


guybedford:
Amazing, thank you.

PR #438, opened 2015-12-23
fix #436: make require() available to config code 

### LINKED ISSUE #436: Regression: 0.14.12 Builder in loadConfigSync require is no longer defined or shimmed?
Greets Guy et al...

I've been updating several of my demo repos and noticed a regression or at least a change that I could not see in any of the code commits tracked in the release notes between JSPM 0.16.15 and 0.16.16. when creating SFX bundles. Example of including extra config loading below:

```
  var builder = new jspm.Builder('./config.js');
  builder.loadConfigSync(extraConfig);
```

Previously when loading that `extraConfig` file which has a `System.config({})` in addition to usage of `require` to load a NPM module before the `System.config` block there was a likely shimmed `require` function available. Print out of the function in 0.16.15 is this:

```
function require(path) {
  return self.require(path);
}
```

It's no longer available in 0.16.16+ when invoking `loadConfigSync`. I assume this associated with SystemJS Builder `0.14.11` -> `0.14.12`. I just couldn't find a change that would remove this functionality. Do you have any insight on this change?

Here is the `extraConfig` file being loaded by loadConfigSync which fails now:
https://github.com/typhonjs-demos/backbone-parse-es6-todos/blob/master/config/config-app-paths.js#L9

fictitious:
another tweak to executeConfigFile

require requires special handling because it's not accessible as global.require - "require isn't actually a global but rather local to each module" as they say at https://nodejs.org/api/globals.html#globals_require

well it will belong to global in the config file execution context because I see no other way around it.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -258,7 +258,7 @@ Builder.prototype.clearCache = function() {
   this.setCache({});
 };
 
-function executeConfigFile(saveForReset, ignoreBaseURL, source) {
+function executeConfigFile(saveForReset, ignoreBaseURL, configPath, source) {
   var self = this;
 
   // create a safe loader to give to the config execution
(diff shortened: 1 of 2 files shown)


fictitious:
travis fails on node 0.10 only, everything newer is OK. Will look into it later.

lib/builder.js
```diff
@@ -273,6 +273,7 @@ function executeConfigFile(saveForReset, ignoreBaseURL, source) {
   var context = Object.create(global);
   context.System = configLoader;
   context.global = context.GLOBAL = context.root = context;
+  context.require = require;
```
guybedford: Strictly speaking, this should probably be a `require` scoped to the config file path itself and not the SystemJS Builder path as the local require here would be.

I think this can be done something along the lines of - `var Module = require('module'); var m = new Module(configPath); var require = m.require.bind(m)`?

guybedford:
Thanks, happy to consider this. Any ideas what the Node 0.10 issue is?

fictitious:
On node 0.10, the `require` itself works, it's the way I'm checking for it in the test that fails. I'm seeing weird thing when the object created by `Object.create(global)` and passed as context to `vm` module seems to lose all properties except `GLOBAL`, `root` and `require`. If I do `var context = {}` and copy all properties from `global` using `for` loop it seems to work.

This is probably just a consequence of `vm

guybedford:
Sure, I see nothing wrong with conditionally disabling the test here then in Node 10.

lib/builder.js
fictitious: Yes, I did that. However node would not search in node_modules, unless m.path is assigned. I had to call Module._nodeModulePaths() - just copied two lines from module.load() code. (the second one assigns m.filename just in case).

guybedford:
Many thanks for your work on this.

PR #439, opened 2015-12-25
Fix Trace getLoadRecord error recovery

### LINKED ISSUE #433: Invalidating file with syntax error breaks further invalidations for that file
If you try to invalidate file with syntax error, and babe failed to transpile it,
if you fix it and tries yo invalidate, you will get the same error.
I believe the builder caches the broken file, and stuck with it because of the babel exception.

Bnaya:
If the getLoadRecord returned promised rejected (like when transpiler exception from syntax error in the module)
the reference to that (rejected)promise kept cached in self.tracing
and further calls to getLoadRecored for that module returned the promise cached in self.tracing

That's breaks continuous builds (tools like jspm-dev-builder)  

fixes jackfranklin/jspm-dev-builder#15
fixes jackfranklin/jspm-dev-builder#11
fixes #433

### DIFF
--- a/lib/trace.js
+++ b/lib/trace.js
@@ -35,7 +35,7 @@ Trace.prototype.traceModule = function(moduleName, traceAllConditionals, conditi
   var loader = this.loader;
 
   var self = this;
-  
+
   return Promise.resolve(loader.normalize(moduleName))
   .then(function(_moduleName) {
     moduleName = getCanonicalName(loader, _moduleName);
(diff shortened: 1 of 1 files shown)


guybedford:
:+1: thanks!

Bnaya:
Thank you!

Bnaya:
what's the next release ETA? (also for jspm)

PR #459, opened 2016-01-11
remove hardcoded traceur version in utils.js

fictitious:
Not sure what the long-term plan is, but traceur main branch is at 0.96 already, so this would need to be updated one way or another


guybedford:
:+1:

Issue #454, opened 2016-01-10
lowResSourceMaps - what does it do?


olsonpm:
> Compile time with source maps can also be improved with the lowResSourceMaps option

I found nothing in the commits nor the code.  I then clone traceur -> become confused as to why this project relies on both traceur and babel -> sustain my confusion as to what lowResSourceMaps does because traceur doesn't explain either.

Please document what lowResSourceMaps does, which in turn should explain how it improves source map compile time

guybedford:
Low res source maps generates lower resolution source maps in Traceur which is faster than the full high resolution source maps it generates by default. I've added a note about this to the API documentation task at https://github.com/systemjs/builder/issues/250#issuecomment-152299085.

olsonpm:
Low Resolution needs to be explained.  Traceur nor its dependency 'mozilla/source-map' explains it.  I guess I'll play around with it tonight and get back with my results.

PR #473, opened 2016-01-18
Clarify lowResSourceMaps option

olsonpm:
- Addresses #454


guybedford:
How did you discover that the mapping granularity was per-line?

PR #482, opened 2016-01-22
Fixes #480 fetch override for `builder.buildStatic`

### LINKED ISSUE #480: Overriding Fetch: buildStatic, doesn't work
I'm trying to adapt the SystemJS Builder (via JSPM, in memory build) to use a custom fetch hook that retrieves file contents from a gulp stream rather than the FS. This works for `builder.bundle`, but not `builder.buildStatic`.

Is there a particular reason for this, or is it just an oversight?

tniswong:
Fixes #480 enabling fetch overrides for `builder.buildStatic`.


guybedford:
Thanks!

Issue #470, opened 2016-01-16
Enhancement: Include "canonicalPath" in external dependencies for SFX bundles. (utils.js->getAlias)


typhonrt:
Greets Guy et al,

So, in my further library writing regarding [backbone-parse-es6](https://github.com/typhonjs-parse/backbone-parse-es6) to solve a serialization issue I now have to override a Parse library file (ParseObject). This new class [BackboneParseObject](https://github.com/typhonjs-parse/backbone-parse-es6/blob/master/src/BackboneParseObject.js#L3-L4) imports a file from `parse/lib/browser/encode.js` with `parse` being aliased to `npm:parse`. When creating a SFX bundle for backbone-parse-es6 which excludes external dependencies including Parse I use this ['meta' config](https://github.com/typhonjs-parse/backbone-parse-es6/blob/master/config/bundle-config.json#L4-L21).

In particular I mark as external dependencies anything under this path: `"parse/lib/browser/*": { "build": false }`

Currently in [utils.js -> getAlias](https://github.com/systemjs/builder/blob/master/lib/utils.js#L60-L85) only the canonical name of the module is resolved which is just `parse` and the path is dropped. This is no good and causes things to break for the SFX bundle. For instance for the AMD bundle it looks like this:
```
(function(factory) {
  define(["underscore","parse","jquery","parse"], factory);
});
```

when it really should be this:
```
(function(factory) {
  define(["underscore","parse","jquery","parse/lib/browser/encode"], factory);
});
```

The potential issue is in [utils.js -> getAlias](https://github.com/systemjs/builder/blob/master/lib/utils.js#L60-L85) and in particular [right here](https://github.com/systemjs/builder/blob/master/lib/utils.js#L76-L79).

Here is a new getAlias with my changes to support canonical paths in addition to canonical name substitution / matching:
```
exports.getAlias = getAlias
function getAlias(loader, canonicalName) {
  var pluginIndex = loader.pluginFirst ? canonicalName.indexOf('!') : canonicalName.lastIndexOf('!');

  if (pluginIndex != -1)
    return getAlias(loader, canonicalName.substr(0, pluginIndex)) + '!' + getAlias(loader, canonicalName.substr(pluginIndex + 1));

  if (canonicalName.match(/\#[\:\{\?]/))
    throw new Error('Get alias not implemented for conditional name "' + canonicalName + '". Remove the conditional exclusion, or post a SystemJS Builder bug if needed!');

  var canonicalPath;

[...]

PR #483, opened 2016-01-24
Fix for Builder Issue #470

typhonrt:
This PR adds support for canonical name + path resolution when a dependency + path is excluded (`build: false`) in an SFX build.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -63,25 +63,36 @@ function toFileURL(path) {
   return 'file://' + (isWin ? '/' : '') + path.replace(/\\/g, '/');
 }
 
-exports.getAlias = getAlias
+exports.getAlias = getAlias;
 function getAlias(loader, canonicalName) {
   var pluginIndex = loader.pluginFirst ? canonicalName.indexOf('!') : canonicalName.lastIndexOf('!');
+
   if (pluginIndex != -1)
     return getAlias(loader, canonicalName.substr(0, pluginIndex)) + '!' + getAlias(loader, canonicalName.substr(pluginIndex + 1));
 
   if (canonicalName.match(/\#[\:\{\?]/))
-    throw new Error('Unable to alias the conditional dependency "' + canonicalName + '".');
+    throw new Error('Get alias not implemented for conditional name "' + canonicalName + '". Remove the conditional exclusion, or post a SystemJS Builder bug if needed!');
+
+  var canonicalPath;
+
+  if (canonicalName.indexOf('/') >= 0)
+    canonicalPath = canonicalName.substr(canonicalName.indexOf('/'), canonicalName.length);
 
   var bestAlias;
 
   function getBestAlias(mapped) {
     return canonicalName.substr(0, mapped.length) == mapped &&
-        (canonicalName.length == mapped.length || canonicalName[mapped.length] == '/');
+     (canonicalName.length == mapped.length || canonicalName[mapped.length] == '/');
   }
 
   Object.keys(loader.map).forEach(function(alias) {
     if (getBestAlias(loader.map[alias]))
+    {
       bestAlias = alias;
+
+      if (canonicalPath)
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -63,25 +63,36 @@ function toFileURL(path) {
...
   if (pluginIndex != -1)
     return getAlias(loader, canonicalName.substr(0, pluginIndex)) + '!' + getAlias(loader, canonicalName.substr(pluginIndex + 1));
 
   if (canonicalName.match(/\#[\:\{\?]/))
     throw new Error('Unable to alias the conditional dependency "' + canonicalName + '".');
 
+  var canonicalPath;
+
+  if (canonicalName.indexOf('/') >= 0)
+    canonicalPath = canonicalName.substr(canonicalName.indexOf('/'), canonicalName.length);
```
guybedford: This isn't quite right. map config can include `/` separators in the alias, so we need to do the path handling after the matching process.

lib/utils.js
```diff
@@ -63,25 +63,36 @@ function toFileURL(path) {
...
   function getBestAlias(mapped) {
     return canonicalName.substr(0, mapped.length) == mapped &&
-        (canonicalName.length == mapped.length || canonicalName[mapped.length] == '/');
+     (canonicalName.length == mapped.length || canonicalName[mapped.length] == '/');
   }
 
   Object.keys(loader.map).forEach(function(alias) {
     if (getBestAlias(loader.map[alias]))
+    {
       bestAlias = alias;
```
guybedford: Just `bestAlias = alias + canonicalName.substr(alias.length);` I think?

guybedford:
Thanks, it would be amazing to get a couple of tests on this as well in https://github.com/systemjs/builder/blob/master/test/canonicals.js.

typhonrt:
I've been testing various changes. 

`bestAlias = alias + canonicalName.substr(alias.length);`  does not work as `canonicalName` in the normal case let's say for parse is: `npm:parse@1.6.14`, so it turns into `"parsearse@1.6.14"` and `"parsearse@1.6.14/lib/browser/encode"` of course the other mappings are equally messed up. 

Another potentially nasty corner case is what if two left hand values share the same right hand value. The last left hand value will always be matched as one is sequentially traversing the System.map array.

guybedford:
Ok I think a `aliasedName = bestAlias + canonicalName.substr(map[bestAlias].length)` then at the end?

In terms of preference, the rules for map config specificity are based on only accepting a new match with the specificity of `alias.split('/').length`. When two matches have the same specificity, the first one should win. See the algorithm at https://github.com/systemjs/systemjs/blob/master/lib/core.js#L26.

The `getBestMatch` function seems badly named too at this point. The algorithm can definitely be cleaned up. It's all well-defined stuff though.

There is one other thing here and that is that this matching is supposed to happen in canonical space for completeness. I'm not too worried about it for now, but just to mention the idea is that `canonicalName` should be decanonicalized via `loader.decanonicalize(canonicalName)` and the map targets should also be decanonicalized this way, and this is the space we should do the matching in. That will handle cases where people manually map into jspm_packages paths like `map: { jquery: 'jspm_packages/npm/jquery@1.2.3/dist/jquery.js' }`.

guybedford:
Sorry I mean URL space not canonical space in the last paragraph.

typhonrt:
This seems to work for the bestAlias return path:

```
  if (bestAlias)
    return bestAlias + canonicalName.substr(loader.map[bestAlias].length);
```

It's pretty simple too. I'll certainly have to defer to you understanding regarding the rest of your comment about the inner workings.

guybedford:
Sure, happy to go ahead with just that fix for now.

typhonrt:
I got lazy and created a new PR: https://github.com/systemjs/builder/pull/485  

Please ignore this one.

Issue #478, opened 2016-01-20
Enhancement: UMD SFX bundle fails w/ Builder 0.15.4 & `meta build false`


typhonrt:
Greets Guy et al,

I haven't had time to track down why there is a failure, but here is what I'm seeing. I have tested between Builder `0.15.3` and `0.15.4` and when trying to bundle a UMD module which uses:
 `"meta": { "underscore": { "build": false } }`

to exclude underscore from the bundle here is the error I receive:
`TypeError: Global SFX bundle dependency "underscore" must be configured to an environment global via the globalDeps option.`

It should be noted that this is a specific change to `0.15.4` as `0.15.3` works fine for creating UMD bundles. It also should be noted that this error doesn't occur for `amd` or `cjs` bundles with the same exact meta / config setup with `0.15.4`. 

I'll try and poke around in the next day or two to find the error, but do post here if you get to it first. 

Thanks.

typhonrt:
@guybedford  So your change in this commit added `umd` bundles to always check that globalDefs is defined if not resolved:
https://github.com/systemjs/builder/commit/496498abdf265010f17581b22440ea5cb75a5ecb#diff-71a1f6779f4a9a1fe69fcec81f85db12R368

IMHO I think UMD bundling should complete and if globalDeps is not available for a given dependency then just the AMD / CJS bundles work in the UMD template.  If globalDeps is defined then AMD / CJS / Global resolve. IE don't enforce the necessity of providing for the Global path. Perhaps a warning on bundling can note that just the AMD / CJS path is resolved. 

I've tested the following replacement for `externalDepIds` resolution in `wrapSFXOutputs` which outputs a warning when there are any non-resolved dependencies from globalDeps:
```
  var externalDepIds = externalDeps.map(function(dep) {
    if (compileOpts.format == 'global' || compileOpts.format == 'umd') {
      var alias = getAlias(loader, dep);
      var globalDep = compileOpts.globalDeps[dep] || compileOpts.globalDeps[alias];
      if (!globalDep)
      {
        if (compileOpts.format == 'umd') {
          console.warn('SystemJS Builder UMD bundle will not support `Global` format as dependency "' + alias  +
           '" is not configured to an environment global via the globalDeps option.');
        }
        else {
          throw new TypeError('Global SFX bundle dependency "' + alias +
           '" must be configured to an environment global via the globalDeps option.');
        }
      }
      globalDeps.push(globalDep);
    }

    // remove external deps from calculated entry points list
[...]

PR #484, opened 2016-01-24
Fix for SystemJS Builder Issue #478

typhonrt:
This PR only will check UMD global compile options if globalName or globalDeps are defined as Builder options.


guybedford:
:+1:

PR #503, opened 2016-02-18
Fix bad object reference

MeoMix:
inlinedLoad has depMap property. inlinedModule is just a string name.


guybedford:
Amazing, thank you for catching this.

davis:
sweet!

mikz:
Yep. I could not make a build. With this fix it works. Thanks!

PR #509, opened 2016-02-23
esm plugins are now correctly transpiled in the legacy path (fixes systemjs/plugin-text#28)

AdamBurgess:
the `source` variable was the one that used to be sent through the transpiler, but it seems a later update went back to using load.source...
(hey, don't linters have 'unused variable' detection?)

### DIFF
--- a/compilers/esm.js
+++ b/compilers/esm.js
@@ -146,7 +146,7 @@ exports.compile = function(load, opts, loader) {
       lowResolutionSourceMap: opts.lowResSourceMaps
     });
 
-    var tree = load.metadata.parseTree || compiler.parse(load.source, load.path);
+    var tree = load.metadata.parseTree || compiler.parse(source, load.path);
 
     if (opts.normalize) {
       var transformer = new TraceurImportNormalizeTransformer(function(dep) {


compilers/esm.js
```diff
@@ -101,7 +101,7 @@ exports.compile = function(load, opts, loader) {
       lowResolutionSourceMap: opts.lowResSourceMaps
     });
 
-    var tree = load.metadata.parseTree || compiler.parse(load.source, load.path);
+    var tree = load.metadata.parseTree || compiler.parse(source, load.path);
```
gholtslander-va: Where does `source` come from? It isn't being passed into this function?

AdamBurgess: scroll up!
https://github.com/AdamBurgess/builder/blob/fix-esm-plugins/compilers/esm.js#L135-L136
it was added in f0236 but then later on was forgotten :(

gholtslander-va: In that case, did you change the wrong one then? You should change this one right? https://github.com/systemjs/builder/pull/509/files#diff-201f308f4e4951a0b727bbb5641b2e3eR149

gholtslander-va: (I may be totally out to lunch here)

gholtslander-va: Actually, changing the one on line 149 fixed this for me locally, whereas changing this one did not.

AdamBurgess: oh, dang, you're right

problem is I work on the systemjs-builder in my node_modules folder then cloned this repo and haphazardly copied the changes onto it

gholtslander-va: Haha. Better change it! It's all good, as I was able to fix this for myself locally to continue development until the patch gets released upstream :sweat_smile:

gholtslander-va:
LGTM :)

AdamBurgess:
thanks for the help @gholtslander-va :palm_tree:

gholtslander-va:
:gift_heart:

guybedford:
:+1:

Issue #515, opened 2016-03-03
When doing SFX build with globalDeps, the output files are not deterministic


mikeyoon:
I noticed when doing an SFX build where I want to exclude certain dependencies, I need to use the globalDeps option to inject them. The order in which those dependencies are injected seem to be non-deterministic due to some object to array conversions.

Here's an example gulp except of what I'm doing.

```
var appVendorExclude = [
  'angular',
  'angular-ui-router',
  'angular-sanitize',
  'lodash',
  'jquery'
];

builder.buildStatic('src/core-app/bootstrap.ts - ' + appVendorExclude.join(' - '), './build/app.js', {mangle: prod, minify: prod, sourceMaps: false,
    globalDeps: {
      angular: 'angular',
      "angular-sanitize": "angular",
      "angular-ui-router": "angular",
      jquery: '$',
      lodash: '_'
    }
  });
```

So output can end up like this:

```
< (["5a","5a"], ["3","18","c","37"], function($__System) {
---
> (["5a","5a"], ["3","18","37","c"], function($__System) {
62714c62714
<     define(["jquery","lodash","angular","angular-sanitize"], factory);
---
>     define(["jquery","lodash","angular-sanitize","angular"], factory);
62717c62717
<     module.exports = factory(require("jquery"), require("lodash"), require("angular"), require("angular-sanitize"));
---
>     module.exports = factory(require("jquery"), require("lodash"), require("angular-sanitize"), require("angular"));
```

I have a simple fix in mind I will make a PR for.

guybedford:
:+1:

guybedford:
Fixed in https://github.com/systemjs/builder/commit/08f69b90a11c24c45f6b1562ab6e440a5ee7dab9.

guybedford:
Released in 0.15.11.

PR #516, opened 2016-03-03
Sort externalDeps when wrapping SFXOutputs

mikeyoon:
Fix for #515.


guybedford:
Thanks, I think a better fix though would to be ensure we use the ordered modules list when extracting the dependencies, since we have an ordered modules array in compile.

Issue #297, opened 2015-09-02
Supporting input sourceMap comment parsing


guybedford:
Parsing out `sourceMap` from input sources and supporting them for the full bundle build. This should go along with https://github.com/systemjs/systemjs/issues/638.

`load.metadata.sourceMap` should obviously still take preference, with these comments effectively filling that meta in SystemJS with the detection enabled.

Similarly, we should strip the comments (along with other unnecessary meta) when doing concatenation.

Rush:
I gave it a shot ... I modified `trace.js` and its source loading step to:
```js
    return Promise.resolve(loader.fetch({ name: normalized, metadata: load.metadata, address: load.address }))
      .then(function(source) {
        
        var sourceMappingURL = require('source-map-url');
        var dataUriToBuffer = require('data-uri-to-buffer');
        var sourceMap = sourceMappingURL.getFrom(source);
        if(sourceMap) {
          source = sourceMappingURL.removeFrom(source);
          if(!load.metadata.sourceMap) {
            load.metadata.sourceMap = dataUriToBuffer(sourceMap).toString('utf8');
          }
        }    
        load.originalSource = source;
        return loader.translate({ name: normalized, metadata: load.metadata, address: load.address, source: source });
      })
```

I can see the final source map's size being affected a little bit and I see that AMD module is reading this source map ... but still it doesn't work. Sources in Chrome show the source after the initial transformation with the inline source map. Any tips? Is it the right place I am tinkering with?

guybedford:
@Rush thanks for looking into this, this looks great, and the right place to be doing it as well. I think this is just about debugging at this point. I'd suggest not setting `originalSource` though as that is for internal use with the compilers. Are you running minification here at all?

Haemoglobin:
Hi,

Just wondering if this issue is related to a problem I'm trying to figure out. 

 I am wanting to go from typescript files with es6 module syntax and transpile those to systemjs format including embedded sourcemaps back to ts files (so far so good, tsc supports this ok). 

Now I want to bundle these into one file by running systemjs-builder on those transpiled files (with sourcemaps turned on). The final sourcemap for the bundle however ends up pointing to transpiled ts source as opposed original ts files. Would this feature allow for this? Or is it possible already?

Thanks!
Hamish

guybedford:
@Haemoglobin yes this sounds like the same issue.

drieks:
Hello, we have the same problem here.
Has anybody new informations about this problem?

[...]

PR #530, opened 2016-03-19
Load inline and external sourcemaps

asapach:
As mentioned in #297, existing sourcemaps are currently being ignored. This should cover the most common scenarios.

Could use some help with testing and verifying that it doesn't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "url": "http://github.com/systemjs/builder.git"
   },
   "scripts": {
-    "test": "mkdir -p test/output && ./node_modules/.bin/mocha"
+    "test": "./node_modules/.bin/mocha"
   },
   "author": "",
   "license": "MIT",
--- a/test/test-build.js
+++ b/test/test-build.js
@@ -1,6 +1,7 @@
 var Builder = require('../index');
 var inline = require('../lib/output').inlineSourceMap;
 var fs = require('fs');
+var path = require('path');
 var spawn = require('child_process').spawn;
 if (process.argv[2] == 'typescript')
   global.ts = require('typescript');
@@ -17,7 +18,7 @@ var builder = new Builder('test/fixtures/test-tree', 'test/fixtures/test-tree.co
 
 function testPhantom(html) {
   return new Promise(function(resolve, reject) {
-    spawn('node_modules/.bin/mocha-phantomjs', [html], { stdio: 'inherit' })
+    spawn('node_modules/.bin/mocha-phantomjs' + (process.platform.match(/^win/) ? '.cmd' : ''), [html], { stdio: 'inherit' })
     .on('close', function(code) {
       if (code !== 0)
         reject(Error('Phantom test failed ' + html + ' failed.'));


guybedford:
This looks really solid. I'm just wondering how well it matches up to the SystemJS handling as the goal is for the builder loading to entirely match the browser loading. 

In particular I'm wondering how much of this behaviour should be handled in the browser loader. While completely unperformant, the nature of the browser loader is already to provide unlinked development performance. So dynamically loading source maps in the browser loader may be worth considering here (perhaps with a meta option to opt-in / out), to improve development source maps (https://github.com/systemjs/systemjs/issues/638) and ensure we have matching behaviours.

If we did that, this code could effectively be moved into the browser loader (the buffer and detection stuff is all translatable). Performance is then the big concern. What do you think?

asapach:
I guess it's theoretically possible to parse inline sourcemaps in the browser, given there's an efficient way to convert base64 data to JSON (my guess is that `btoa` will choke on large sourcemaps). `ArrayBuffer` or `Buffer` polyfill?

External sourcemaps are more tricky - they require an XHR call. How do you abstract this away so that it would make an `fs.readFile()` during build and an XHR in the browser?

guybedford:
`System.fetch` provides a read abstraction. 

The main browser cases here are:
* Loading a file that has an inline source map comment, and then running that as CommonJS, global, AMD or System.register shouldn't actually need to parse the comment because the browser will get it and these are non-transforming formats for most purposes.
* Loading a file that has an inline source map comment that then goes through a plugin transformation step. Here if it has an inline source map comment it would be nice to support that through the plugin transformation. Loading the source map would incur additional latency, but that is probably expected. The implementation here would be exactly like this PR, but in the browser. The only thing is this is likely not a very common case to put a lot of work into supporting.

Thinking about this further, I guess we only really need the first case above to work, and that is particularly what issue systemjs/systemjs#638 is tracking. The second case is very much an edge case and something that does seem highly unnecessary to support. The first case just involves detecting the source map comment, and ensuring it remains detectable by the browser.

I do still think we should aim for complete parity with the browser scenario for completeness. So if in the browser we are going to not support inline source maps for translated files then we shouldn't during the builder either.

That is, I do think this would be best moved to the `lib/output.js` of `lib/sourcemap.js` file handling, as that way the plugin input will not already have source maps parsed to provide this exact parity.

Let me know if that makes some sense. The parity point may be taken too far here, but I think it's a nice implementation principle to try and stick to even for cases like this.

asapach:
> Loading a file that has an inline source map comment, and then running that as CommonJS, global, AMD or System.register shouldn't actually need to parse the comment because the browser will get it and these are non-transforming formats for most purposes.

By non-transforming you mean that it should be safe to modify the first (and last) line without affecting the sourcemap? That might not always be the case - here's an example [link](https://cdnjs.cloudflare.com/ajax/libs/backbone.js/1.3.2/backbone-min.js). If you add an IIFE wrapper on the first line you will render the sourcemap unusable.

> I do still think we should aim for complete parity with the browser scenario for completeness. So if in the browser we are going to not support inline source maps for translated files then we shouldn't during the builder either.

It's very frustrating that in this case it's impossible to use `systemjs-builder` as a build step in a chain. For example I first run `babel` on every individual file (which also produces the sourcemaps) and then I would like to run the builder as the next step, but I can't because the sourcemaps are being ignored. I have to effectively transpile twice: first each individual file and then using the builder passing the `transpiler` option.

guybedford:
There's some misunderstanding here. By translated source I mean source translated by the SystemJS translate hook. That is chaining the input source map with further compilation steps should not be supported, but loading an input with a source map should be supported. And the way to do this correctly is not through the fetch hook here, but through the output.js or sourcemap.js file to ensure we handle this _after_ transpilation, not _before_.

To explain the implementation detail here in SystemJS - when it does wrapping like with CommonJS and globals, it will never add a newline to the leading wrapper - `module.exports = 'x';` -> `(function(require, module, exports) { module.exports = 'x';` so we maintain the line offsets that the source map uses, and only lose the ability to get line debugging on the very first line of code.

asapach:
> so we maintain the line offsets that the source map uses, and only lose the ability to get line debugging on the very first line of code.

That's what I'm trying to illustrate: if you use a minified file (like my backbone example) the whole file may be in the one first line.

guybedford:
Right, that's the edge case certainly, but most projects will have a license comment even when minified. Also it's advisable to use unminified sources for debugging. It's a browser-specific edge case though, so I'm not sure how it relates to the discussion here?

PR #538, opened 2016-03-31
fix unit tests for windows

arackaf:
I'll be perfectly honest: I'm not entirely certain WHY the switch to child_process.spawnSync fixed most of my issues.  But it did.

The fix to `Test AMD format` seems a bit more straightforward: the prior phantom tests do appear to have modified the builder instance's configuration.  The only thing not clear to me is why it doesn't fail for everyone.

### DIFF
--- a/test/test-build.js
+++ b/test/test-build.js
@@ -2,7 +2,7 @@ var Builder = require('../index');
 var inline = require('../lib/output').inlineSourceMap;
 var fs = require('fs');
 var path = require('path');
-var spawn = require('child_process').spawn;
+var spawnSync = require('child_process').spawnSync;
 if (process.argv[2] == 'typescript')
   global.ts = require('typescript');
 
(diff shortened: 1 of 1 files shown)


arackaf:
Well damnit I broke Node 0.10.  You sure you need to support 0.10 :-) 

But seriously, hopefully this is of some use to you - hopefully there's some tweak you can make to fix both my failures, and the CI tests.

guybedford:
It's difficult to move forward here without being able to replicate the issue at all. Which shell are you using in windows?

arackaf:
Windows 10 with a plain-old cmd window.  npm run test.  That's all, I'm afraid.  Commenting out those tests isn't a huge deal when working here, if you can't repro those failures.

guybedford:
Perhaps https://github.com/systemjs/builder/pull/557/files#diff-62ce57f12037115915b4626a5c623234R21 is a suitable fix here?

Issue #555, opened 2016-04-18
Babel incorrectly names dependencies that have a trailing !


npbenjohnson:
When a resource is loaded via trailing ! plugin syntax, the module uses a mapped name, and dependencies use the name as is. Trailing ! syntax is the only way to get the systemjs-hot-reloader to detect changes in files imported with plugins.

guybedford:
Can you clarify the exact issue here a little? Are you saying that you are using `resource.x!` plugin syntax through systemjs-hot-reloader to get plugin hot reloading support, but then Babel will incorrectly name them in bundles breaking the build?

What name is Babel outputting? Adjusting the normalization is unfortunately not the fix though as the name should remain unnormalized as-is for normal (non normalizing) bundling.

npbenjohnson:
If you bundle 2 files [a.js, b.txt!], and a references b, the register statements look like:
```
System.register('a.js', ['b.txt!'] ... )
System.register("b.txt!text.js" ... )
```
so when you load a, it tries to find b externally

guybedford:
The reason for this is that bundles don't include normalization. It's badly documented yes, but bundles are supposed to be added after including the original configuration. Bundles are only a file-level multiplexing and do not build in linking information.

npbenjohnson:
That makes sense, buildStatic covers what I actually needed for this use case.

PR #556, opened 2016-04-18
babel correctly bundles dep names ending with !

npbenjohnson:


npbenjohnson:
https://github.com/systemjs/builder/issues/555

npbenjohnson:
bad checkin, missing a return statement

Issue #559, opened 2016-04-19
Static/SFX build errors in browser when commonjs module exports window.location as default export


joeldenning:
This issue is basically identical to systemjs/systemjs#1192 (fixed with systemjs/systemjs#1193), except the problem is manifested in sfx builds. Basically, IE11 throws an Error if you call `Object.getOwnPropertyDescriptor` on `window.location`.

A popular npm module that does this is [sockjs-client](https://github.com/sockjs/sockjs-client), specifically their [location](https://github.com/sockjs/sockjs-client/blob/master/lib/location.js#L3) module.

I've got a PR coming that should fix this, with code being the same as what I did in systemjs/systemjs#1192

guybedford:
Released in 0.15.16.

PR #560, opened 2016-04-19
Fixing #559 - runtime error when commonjs module.exports is window.location

joeldenning:
See #559, systemjs/systemjs#1192, and systemjs/systemjs#1193. I literally copied and pasted the code from the systemjs pull request into the two templates here in the builder. I verified locally that my project (that was failing to bootstrap in the browser because of the problem) is now successfully bootstrapping.

Let me know if you'd prefer me to take the minified/built files out of this pull request. Also, I'm not sure what the `.uin.js` files are but I didn't see a way to rebuild those files.

### DIFF
--- a/templates/sfx-core-register.js
+++ b/templates/sfx-core-register.js
@@ -139,18 +139,16 @@
     var esModule = {};
     // don't trigger getters/setters in environments that support them
     if (typeof exports == 'object' || typeof exports == 'function') {
+      var hasOwnProperty = exports && exports.hasOwnProperty;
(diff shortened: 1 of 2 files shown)


joeldenning:
I decided that you probably didn't want me committing the minified files, so I took them out of this pull request. Let me know if you want me to reinstate them, although I'd need instructions on what to do to build the `.uin.js` files.

guybedford:
Thanks, yes this is just a direct copy here. I'll run the minify build for publish.

Issue #561, opened 2016-04-19
Output minification step breaks SourceMap chain with inlined `sourceContents`


peta:
Problem:
===
When I enable output minification and have enabled inlined sources (called `sourceContents` and enabled by setting either the builder option `sourceMapContents` to true or setting `sourceMaps` to `'inline'`) the final SourceMap generated for my build does no longer contain them.

Reason:
===
When `minify: true` the final minification step in `lib/output#writeOutputs` uses `uglify-js` for source minifcation. But in there neither the uglify option `sourceMapIncludeSources` is set (don't know if this is mandatory but it is told in the official uglifyjs docs) nor are the sourcesContent items passed to the `uglify.SourceMap` instance by calling `setSourceContent` for them.

guybedford:
👍

guybedford:
Released in 0.15.16.

PR #562, opened 2016-04-19
Fix: Output minification step breaks SourceMap chain with inlined `sourceContents` #561

peta:
Addresses issue #561


guybedford:
Amazing work, thank you so much for putting together the PR.

PR #568, opened 2016-04-25
added support for input source maps

asapach:
The change adds an extra step before compilation to read the input source map (either inline or in a file). Added some example files with source maps and a unit test to verify.

Any suggestions on refactoring this are welcome.


guybedford:
Thanks for this work, I will review as soon as I can.

As a side note, did you try checking the validity of the outputs here? I've been doing some testing related to the other issues by investigating source maps in https://github.com/sokra/source-map-visualization, and am getting a lot of errors. As far as I can tell this seems to be directly related to Traceur's handling of input source maps.

asapach:
I'm using WebStorm - it has a built-in source map visualizer (although I can't find it mentioned in the docs anywhere to point you to) - when you open a `.map` file it gives you a split view of the transpiled lines on the left mapped to source lines on the right. You can click around and see the mappings at work.

I've also tried loading the files in the browser and poking around in devtools - when you set a breakpoint it jumps from the transpiled code to the source line - but this gets pretty tedious.

I have indeed noticed some inconsistencies in the ouput source maps: sometimes the mappings are off by several characters, or some statements get merged together, or by contrast one statement gets split into several mappings without obvious reason. And in some cases the maps are completely garbled and unusable. I can come up with some examples if that helps.

As far as I can tell it happens when you use `transpiler` option as well, so my changes don't affect this either way.

lib/compile.js
```diff
@@ -105,6 +106,23 @@ function compileLoad(loader, load, compileOpts, cache) {
     if (format == 'cjs')
       mappedLoad.source = mappedLoad.source.replace(hashBangRegEx, '');
     return Promise.resolve()
+    .then(function () {
+      var sourceMap = sourceMappingURL.getFrom(mappedLoad.source);
```
guybedford: If the `metadata.sourceMap` is already set, we should probably skip all of this checking? Or do we still want to support that case as well?

guybedford:
This looks great to me, happy to merge!

Any chance we can rebase to the latest master?

lib/compile.js
guybedford: I think skipping on `metadata.sourceMap` already set makes the most sense. Plugins should handle extracting sourceMappingURL comments themselves if they want to support transitivity, as we can't just "add it on" here for a plugin that has already transpiled.

asapach: Sure, will fix it.

asapach:
Sure, I wanted to run some more tests on real-life projects first. I'll update the PR tomorrow.

asapach:
@guybedford, rebased and fixed the review issues.
Also did some real-life tests and it looks like `gulp-sourcemaps` is broken, because the `file` and `sources` it sets are not relative to the output file, but rather to the base. And `sourceRoot` is either invalid (e.g. hardcoded `/source/`) or useless. Something to keep in mind.

asapach:
Experimentally figured out the right setup for `gulp-sourcemaps`, here's an example for typescript:
```js
    return gulp.src('src/ts/**/*.ts', { base: '.' }) // base here is important, since it needs to match `sourceRoot`
        .pipe(sourcemaps.init())
        .pipe(ts(...))
        .pipe(sourcemaps.write({ sourceRoot: '.' })) // by default `sourceRoot` is `/source/`, which makes no sense
        .pipe(gulp.dest(dist));
```

asapach:
also made sure the `sourceMappingURL` comment is removed from all the sources, including the ones embedded in `sourceMapContents`

lib/compile.js
```diff
@@ -105,6 +106,26 @@ function compileLoad(loader, load, compileOpts, cache) {
     if (format == 'cjs')
       mappedLoad.source = mappedLoad.source.replace(hashBangRegEx, '');
     return Promise.resolve()
+    .then(function () {
```
guybedford: I think we should move this whole implementation to be an extension of the loader itself, applied in the fetch hook at https://github.com/systemjs/builder/blob/master/lib/builder.js#L195. This way, plugins will be able to respond to input sourceMaps during builds if they wish to as well.

guybedford: The only issue with this approach is if the fetched source is not JavaScript, but I think the source map detection can work equally across input types fine so this won't be a bad failure, worst case we return `metadata.sourceMap` when we shouldn't. But we should ensure this process itself is resilient to errors at least.

lib/compile.js
```diff
@@ -105,6 +106,26 @@ function compileLoad(loader, load, compileOpts, cache) {
...
+      var metadata = mappedLoad.metadata;
+      if (metadata.sourceMap)
+        return;
+      var sourceMap = sourceMappingURL.getFrom(mappedLoad.source);
+      if (!sourceMap)
+        return;
+      mappedLoad.source = sourceMappingURL.removeFrom(mappedLoad.source);
+      if (sourceMap.startsWith('data:')) {
+        // inline source map
+        metadata.sourceMap = JSON.parse(dataUriToBuffer(sourceMap).toString('utf8'));
```
guybedford: Perhaps we should wrap this with a try-catch to avoid throwing the build if it's wrong?

lib/compile.js
```diff
@@ -105,6 +106,26 @@ function compileLoad(loader, load, compileOpts, cache) {
...
+      } else {
+        // relative path to the .map file
+        var sourcePath = path.resolve(fromFileURL(loader.baseURL), load.path);
+        var sourceMapPath = path.join(path.dirname(sourcePath), sourceMap);
+        return asp(fs.readFile)(sourceMapPath, 'utf8').then(function (sourceMap) {
```
guybedford: Again, we should add a catch handler if the file is not found.

lib/compile.js
```diff
@@ -105,6 +106,26 @@ function compileLoad(loader, load, compileOpts, cache) {
...
+          metadata.sourceMap = JSON.parse(sourceMap);
```
guybedford: We allow `metadata.sourceMap` to be an object fine, so we should be able to skip the `JSON.parse` here.

guybedford: I think!?

guybedford:
Thanks for your patience here. It looks great to merge to me, just let me know your thoughts on the final feedback.

lib/compile.js
asapach: OK, I'll give it a try.

lib/compile.js
asapach: OK, I'll see if it works.

lib/compile.js
asapach: What's the correct way to report an error here? I assume `console.error()` is not good enough?

guybedford: Unfortunately we don't have a good warning system at the moment, and yes we don't want to just be writing to the console from within the API.

I was thinking of just having a silent error, because source maps fail silently in the browser when they don't work so that seemed like it may be an ok precedent. But let me know if that sounds unreasonable?

asapach: As a developer I hate silent errors. Any plans to address this in the future?

guybedford: Sure, if you like we can leave it as a full thrown error for now, and rather handle this in future when it comes up again.

guybedford:
Let me know if the feedback makes sense here. If it is a problem at all, I'm happy to manually merge the feedback in here too.

lib/compile.js
asapach: Tried this and I get the following error: `load.metadata.sourceMap must be set to an object.` somewhere in `loader.instantiate`

guybedford: Sorry I misread this as a JSON.stringify, ignore this comment entirely then.

asapach:
Sorry, I've been busy with work.
I gave the fetch hook a try and I have a branch if you want to take a look: https://github.com/systemjs/builder/compare/master...asapach:fetch
The problem I ran into is that the following code ruins everything: https://github.com/systemjs/builder/blob/85545097693fd6d51b8d23222277bb877f39d93e/lib/trace.js#L460-L472
It rewrites the source map and changes the file names, so they no longer make sense.

asapach:
OK, I think I found a workaround. @guybedford, care to take a look?

asapach:
@guybedford did you have a chance to take a look at this? Should I rebase to the latest master?

lib/trace.js
```diff
@@ -461,14 +461,16 @@ Trace.prototype.getLoadRecord = function(canonical, excludeURLs, parentStack) {
...
 
-            var originalName = load.name.split('!')[0];
-            
-            // force set the filename of the original file
-            sourceMap.file = originalName + '!transpiled';
-
-            // force set the sources list if only one source
-            if (!sourceMap.sources || sourceMap.sources.length <= 1)
-              sourceMap.sources = [originalName];
+            if (load.name.indexOf('!') != -1) {
```
guybedford: Could you clarify how this workaround helps here? Are you only using input source maps with plugins loaded with `!` syntax? If so it may be too specific a workaround unfortunately?

Perhaps we can just change this code to be non-damaging rather, only setting `file` and `sources` properties that are not already set.

asapach: > Are you only using input source maps with plugins loaded with ! syntax?

Not quite. The input source maps work regardless of this check, but this code changes the file names, which makes source maps useless, so now it only runs for plugins. I'm not sure if this is correct and what was the original intention or how my change will affect this.
I have a couple of examples in this PR:
* `register.js` has source map indicating that `babel.js` is its original source, but this code changes it to `register.js!transpiled → register.js`
* similarly `example.js → example` becomes `example.js!transpiled → example.js`

sergei-startsev:
@guybedford I've came across the same issue with understanding already generated sourcemaps, unfortunately it looks like current `master` branch doesn't include any information about transpiled code without `transpiler` option in systemjs configuration, so you cannot retrieve original code if you decide to build a bundle based on already transpiled code, even if sourcemaps are provided. Actually it confuses. And you also need to transpile code twice.

I've verified @asapach pull request for one of my projects (typescript+systemjs) and it works perfect. Here are examples of gulp tasks that I use:

```javascript
gulp.task('transpile-typescript', ['clean'], function () {
    return gulp.src('src/ts/**/*.ts', { base: '.' })
        .pipe(sourcemaps.init())
        .pipe(ts({
            noImplicitAny: true,
            target: "es5",
            module: 'system',
            allowSyntheticDefaultImports: true
        }))
        .pipe(sourcemaps.write('.', {
            sourceRoot: '.'
        }))
        .pipe(gulp.dest(dist));
});

sergei-startsev:
Before I found that PR as a temporary solution I used [sorcery](https://www.npmjs.com/package/sorcery), that magical package also works fine, but I'm looking forward supporting it out of box. Hopefully the PR will be accepted asap.

asapach:
@guybedford, any suggestions how to best handle this '!transpiled` case?

lib/trace.js
guybedford: Ok, we should just be able to replace this with `if (load.metadata.loader)` rather than plugin syntax specifically, so it applies to all plugins.

I will then also make this change in https://github.com/systemjs/systemjs/blob/master/lib/plugins.js#L202 for the next SystemJS release to match.

Let me know if that works for you?

guybedford:
I think it does make sense that we only "sanitize" the source map in the plugin case, where it should represent the direct original source transformation.

Let me know if you can do that update and rebase one more time, then lets get this done for the release hopefully tomorrow.

lib/trace.js
guybedford: https://github.com/systemjs/systemjs/blob/master/lib/plugins.js#L202 is obviously already filtered to this plugin case...

asapach:
OK, will rebase and make the change in a few minutes.

guybedford:
Thanks!

asapach:
Done.

guybedford:
Excellent, thank you for this amazing work!

asapach:
Thanks for your patience :)

PR #569, opened 2016-04-28
node.js 6 fixes

papandreou:
The enclosed changes make assetgraph's systemjs-builder tests work with node.js 6. Previously it failed like this: https://travis-ci.org/assetgraph/assetgraph/jobs/126310648

The `configPathCache` thing seems a bit weird to me -- when would an object ever have a `map` method?

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -145,7 +145,7 @@ function inlineSourceMap(source, sourceMap) {
 exports.writeOutputs = function(outputs, baseURL, outputOpts) {
   var outFile = outputOpts.outFile && path.resolve(outputOpts.outFile);
   var basePath = fromFileURL(baseURL);
-  var fileName = path.basename(outFile) || 'output.js';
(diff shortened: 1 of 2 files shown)


guybedford:
packageConfigPaths is supposed to be an array, so this may be caused somewhere else in the process, by passing a non-array into system.config here. Perhaps this is happening during a serialization process of the configuration?

I've also added Node 6 testing to the main branch here which needed another fix as well.

papandreou:
> packageConfigPaths is supposed to be an array, so this may be caused somewhere else in the process, by passing a non-array into system.config here. Perhaps this is happening during a serialization process of the configuration?

Doesn't look like any of the tests do that. Hmm...

guybedford:
It seems like you've got this working? Can I close?

papandreou:
Yeah, thanks. I must have screwed up when I `npm linked`ed systemjs-builder in and switched to node.js 6. It's clearly working now :)

PR #595, opened 2016-05-16
Only copying default property of commonjs modules once onto esModule. Fixes jspm#1793

joeldenning:
This fixes jspm/jspm-cli#1793.


guybedford:
Thanks for testing this!

PR #612, opened 2016-05-29
fix(): loader should not fetch sourcemaps if disabled.

### LINKED ISSUE #610: Static build fails due to a missing source map file
After upgrading to `v0.15.18` I get the following error during build:

```
... [SOME LINES ARE OMMITTED] ...
[11:46:47] 'dist:bundle-app-js' errored after 314 ms
[11:46:47] Error on fetch for vendor:es6-shim.js at file:///Users/korya/dev/embedded_fw/recipes-planitar/survey/build/vendor/es6-shim.js
	Loading app/bootstrap.js
	Error: ENOENT: no such file or directory, open '/Users/korya/dev/embedded_fw/recipes-planitar/survey/build/vendor/es6-shim.map'
    at Error (native)
```

***

Indeed `build/vendor/es6-shim.js` references its source map file in a comment:

```js
...
//# sourceMappingURL=es6-shim.map
```

This files does not exist in `build/vendor` because I don't need it. I bundle my application without source maps.

***

Here is my setup. It **worked fine with `v0.15.13`**.

I'm using the following gulp task to bundle my code:

```js
gulp.task('dist:bundle-app-js', ['build'], (cb) => {
  var builder = new systemjsBuilder('./build', 'config.js');

  return builder.buildStatic('app/bootstrap.js', 'www/bundle.js', {
    runtime: false,
    minify: true,
    mangle: true,
  });
});
```

Contents of `config.js`:

```js
System.config({
  baseURL: '/',
  defaultJSExtensions: true,
  paths: {
    'vendor:*': 'vendor/*',
  },

  map: {
    'es6-shim': 'vendor:es6-shim.js',
    'lodash': 'vendor:lodash.js',
    'jquery': 'vendor:jquery.js',
    'jquery-ui': 'vendor:jquery-ui.js',
    'bootstrap': 'vendor:bootstrap.js',
    'angular': 'vendor:angular.js',
    'angular-cookies': 'vendor:angular-cookies.js',
    'angular-touch': 'vendor:angular-touch',
    'angular-ui-router': 'vendor:angular-ui-router',
    'ui-bootstrap': 'vendor:ui-bootstrap',
    'toastr': 'vendor:toastr',
    'hammer': 'vendor:hammer',
    'propagating': 'vendor:propagating',
  },
});
```

DevVersion:
* Currently if the trace option `sourceMaps` is set to false, the sourceMaps will be still loaded.
  This isn't valid. Once the option is set to false, it should just ignore the sourceMaps.

> References https://github.com/angular/material2/issues/541

FYI: I'm not sure, this is the correct way to transmit the trace options.

Fixes #610.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -5,8 +5,6 @@ var asp = require('bluebird').promisify;
 var fs = require('fs');
 var path = require('path');
 
-var dataUriToBuffer = require('data-uri-to-buffer');
-
 var extend = require('./utils').extend;
 
 var attachCompilers = require('./compile').attachCompilers;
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks, we can add this functionality, but yes we should work out how to properly pass the trace options.

It may make sense to move the source maps fetch behaviour back into `lib/trace.js` as the first step after the `loader.fetch` call.

guybedford:
Thanks!

PR #616, opened 2016-06-07
Fixed  #615 - Compile#getEncoding() should consider npm-scope package

### LINKED ISSUE #615: Compile#getEncoding() should consider npm-scope package
According to this line - https://github.com/systemjs/builder/blob/master/lib/compile.js#L46
If the canonical is a [npm-scope package](https://docs.npmjs.com/misc/scope), it won't be encoded.

For example,
```import {MyFunc} from '@myorg/mypackage';```
This will be output as follows:
```
$__System.register("@myorg/mypackage", [], function(exports_1, context_1) {
```
So, it may be possible to check if the canonical contains a slash `/` then it won't be a system module.

jumperchen:
Fixed #615 - Compile#getEncoding() should consider npm-scope package

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -43,7 +43,7 @@ function getCompileHash(load, compileOpts) {
 
 function getEncoding(canonical, encodings) {
   // dont encode system modules
-  if (canonical[0] == '@' && canonical != '@dummy-entry-point')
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks!

PR #626, opened 2016-06-15
Docs update

arackaf:
Add some notes on buildStatic, which should help devs trying to support static builds being loaded from script tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,9 @@ This bundle file can then be included with a `<script>` tag, and no other depend
 to export a global variable when loaded from a script tag, and this can be configured via `globalName`.  For example
 
 ```javascript
-builder.buildStatic('src/NavBar.js', 'dist/NavBarStaticBuild.js', { globalName: 'NavBar' });
+builder.buildStatic('src/NavBar.js', 'dist/NavBarStaticBuild.js', { 
+  globalName: 'NavBar' 
+});
 ```
 
 will cause the output of your module to be assigned to a global variable named `NavBar`.  If you're making a static bundle, while excluding certain dependencies, those dependencies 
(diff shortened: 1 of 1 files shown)


lookfirst:
👍

README.md
```diff
@@ -105,9 +105,34 @@ To make a bundle that is independent of the SystemJS loader entirely, we can mak
...
+```javascript
+builder.buildStatic('src/NavBar.js', 'dist/NavBarStaticBuild.js', { globalName: 'NavBar' });
+```
+
+will cause the output of your module to be assigned to a global variable named `NavBar`.  If you're making a static bundle, while excluding certain dependencies, those dependencies 
+will of course need to have already been loaded on your page, with their own global variables exported.  You can match these global variables up with your needed dependencies 
+with `globalDeps`.  For example
+
+```javascript
+builder.buildStatic('src/NavBar.js - react', 'dist/NavBarStaticBuild.js', { globalName: 'NavBar', globalDeps: { 'react': 'React' } });
```
guybedford: It may be worth documenting `externals` here:

```javascript
builder.buildStatic('src/NavBar.js', 'dist/NavBarStaticBuild.js', {
  externals: ['react'],
  globalName: 'NavBar',
  globalDeps: {
    'react': 'React'
  }
});
```

I also just added https://github.com/systemjs/builder/commit/c3c3e73475ca28e1a7b235ac02338c5d716e0983 so that globalDeps can be treated as externals automatically:

```javascript
builder.buildStatic('src/NavBar.js', 'dist/NavBarStaticBuild.js', {
  globalName: 'NavBar',
  globalDeps: {
    'react': 'React'
  }
});
```

guybedford: The reasoning here is that exclusion does tree exclusion, which is not always wanted as you might lose a dep of `src/NavBar.js` shared with React that you actually wanted in the build.

guybedford:
Excellent thanks so much @arackaf this is a massive help!

arackaf:
@guybedford I'd be happy to add a description of externals to the pr, but I'm afraid I don't really know what that is or how it works :)  

Everything I documented above is just what I either surmised from the error messages, or got help with from you in GH issues.

So, what is externals and how does it work?

guybedford:
@arackaf sure `externals` is exactly like the subtraction above, except it won't remove a shared dependency (say if app and React both depend on fs, you will still bundle fs) and it won't trace the dependency at all so will build faster.

guybedford:
Externals mostly sugars to the `build: false` metadata.

arackaf:
Ah - nice.  Yeah, I'll add that a bit later.  Thanks.

arackaf:
Ok - added a quick note about externals - I didn't actually test that externals works as you described, but if I can't trust *you* on such things.... :)

guybedford:
Thanks!

PR #636, opened 2016-06-28
Make builder accessible in plugins

casperlamboo:
In some plugins the builder has to be accessible, for instance the WebWorker plugin.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -121,12 +121,10 @@ Builder.prototype.reset = function(baseLoader) {
   baseLoader = baseLoader || this.loader || System;
 
   var loader = this.loader = new baseLoader.constructor();
-  loader.builder = true;
+  // make builder accessible in plugins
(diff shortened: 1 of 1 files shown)


lib/builder.js
```diff
@@ -124,6 +124,9 @@ Builder.prototype.reset = function(baseLoader) {
   loader.builder = true;
   // put the loader in the builder environment
   loader.config({ build: true });
+  
+  // make builder accessible in plugins
+  loader.systemJSBuilder = this;
```
guybedford: How about making this `loader.builder`? This does collide with the true value above I know, but checks like `if (this.builder)` won't be affected?

guybedford:
Thanks!

lib/builder.js
casperlamboo: Good point, updated the pr

guybedford:
👍

PR #645, opened 2016-07-12
demonstrate issue with load.metadata.timestamp not being set

frankwallis:
This is a repro for issue (1)  mentioned [here](https://github.com/frankwallis/plugin-typescript/issues/133#issuecomment-231892589) where ```load.metadata.timestamp``` is not being set and so caching is not working.

The calllback method which sets ```load.metadata.timestamp``` is never called in ```builder.loader.fetch``` because of an error [here](https://github.com/systemjs/builder/blob/master/lib/builder.js#L179)

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -176,8 +176,7 @@ Builder.prototype.reset = function(baseLoader) {
   var loaderFetch = loader.fetch;
   loader.fetch = function(load) {
     var self = this;
-    return Promise.resolve((builder.fetch || loaderFetch).call(this, load, function(load) {
-      return Promise.resolve(loaderFetch.call(self, load))
+    return Promise.resolve((builder.fetch || loaderFetch).call(this, load))
       .then(function(source) {
         // calling default fs.readFile fetch -> set timestamp as well for cache invalidation
         return asp(fs.stat)(fromFileURL(load.address))
@@ -190,8 +189,7 @@ Builder.prototype.reset = function(baseLoader) {
             return source;
           throw err;
         });
-      })
-    }));
+      });
   };
 
   // this allows us to normalize package conditionals into package conditionals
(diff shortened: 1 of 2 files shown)


test/test-build-cache.js
```diff
@@ -110,6 +110,21 @@ suite('Test compiler cache', function() {
     assert.deepEqual(invalidated, [builder.loader.normalizeSync('deep/wildcard/test.js')]);
   });
 
+  test('builder.loader.fetch sets load.metadata.timestamp', function() {
+    var source = 'export var p = 5;';
+    var builder = new Builder('test/output');
+    fs.writeFileSync('./test/output/timestamp-module.js', source);
+    var address = builder.loader.normalizeSync('./test/output/timestamp-module.js');
+    var load = { name: address, address, metadata: {} };
```
guybedford: For legacy Node support can we do `address: address` here?

lib/builder.js
```diff
@@ -176,8 +176,7 @@ Builder.prototype.reset = function(baseLoader) {
   var loaderFetch = loader.fetch;
   loader.fetch = function(load) {
     var self = this;
-    return Promise.resolve((builder.fetch || loaderFetch).call(this, load, function(load) {
-      return Promise.resolve(loaderFetch.call(self, load))
+    return Promise.resolve((builder.fetch || loaderFetch).call(this, load))
```
guybedford: I think the fix would be to write it as:

```javascript
var loaderFetch = loader.fetch;
var cachedFetch = function(load) {
  return Promise.resolve(loaderFetch.call(this, load))
  .then(function(source) {
    // ... exact same code setting load.metadata.timestamp here
  });
};
loader.fetch = function(load) {
  if (builder.fetch)
    return builder.fetch.call(this, load, cachedFetch);
  return cachedFetch.call(this, load);
};
```

guybedford:
Well done for spotting this!

frankwallis:
I have updated the PR with your feedback and the tests are now passing. Also thank you for your very kind donation, it is much appreciated!

guybedford:
Excellent, thanks for the update.

Issue #489, opened 2016-01-31
Upgrade to using Babel over Traceur for compiler transformers


guybedford:
All the transformers in https://github.com/systemjs/builder/tree/master/compilers should be upgraded to be based on Babel transformation over Traceur transformations.

This would then support comments in the output source.

guybedford:
In addition this will better support source maps.

dschnare:
Is there anyway to have the `transpiler` config option take effect instead of having the transpiler hardcoded to Babel/Traceur? Using `System.config()` then `System.import()` it respects the custom `transpiler`, but `builder.config()` seems to have difficulty tracing the dependency graph for ES6 modules when I have a custom transpiler specified.

If it helps I was attempting to use the Typescript compiler.

guybedford:
It's much better to use http://github.com/frankwallis/plugin-typescript over the internal TypeScript transpiler.

dschnare:
Thanks, I'll give that a shot.

rjharmon:
I'm left confused by the various tickets related to this.  While I don't see that it's solved, I don't see any tickets still open for solving it.  a) is it actually solved?  b) if so, what package versions would have the improvements, and c) if not, what open issue(s) remain (or does one or more of the existing closed issues need to be re-opened?)?

Thanks!

guybedford:
This is entirely to do with SystemJS builder's internal transformation code and not to do with which ES6 transpiler is being used.

jrauschenbusch:
Hi @guybedford,

can you give a rough estimate when the upgrade to Babel for the Internal transformation code will be completed?

Thanks for your great work on jspm, systemjs and systemjs builder!

guybedford:
@jrauschenbusch it's not currently a priority unfortunately.

asapach:
@guybedford when will this become a priority? The source maps issue is a show-stopper for production use.
Is this something the community can help with?

guybedford:
@asapach the task here is simply to rewrite the files in https://github.com/systemjs/builder/tree/master/compilers as Babel transforms. The tests cover probably about 70% of the cases, with the rest of the details in the code itself, so I would gladly provide feedback if someone wants to take this on before I can get to it.

asapach:
OK, I'll give it a try. Which one is the simplest one to start with in your opinion?

guybedford:
https://github.com/systemjs/builder/blob/master/compilers/register.js is the simplest.

asapach:
[...]

PR #652, opened 2016-07-23
upgraded register transformer to babel transform-system-register

asapach:
First step towards progress in #489


test/sourcemaps.js
```diff
@@ -141,7 +140,7 @@ suite('Source Maps', function() {
           '../fixtures/test-tree/some.js',
           '../fixtures/test-tree/text.txt',
           '../fixtures/test-tree/amd.js',
-          '../fixtures/test-tree/component.jsx',
+          '../fixtures/test-tree/unknown', // component.jsx
```
guybedford: Is this that sourcemaps issue you were talking about? I may be able to see if I can look into this too.

guybedford:
This is awesome! I can't express how grateful I am to see work here moving forward.

I've given you collaborator access now, so if you want to transfer https://github.com/asapach/babel-plugin-transform-system-register across into the SystemJS organization we can do that. Either way is fine.

test/sourcemaps.js
asapach: Yes, I think previously the sourcemap contained `component.jsx → unknown` (but traceur for some reason ignored the latter), but now babel ignores the intermediate file.

asapach:
Here's what I get when I try to transfer:
transfer
Am I doing this right? Maybe you need to fork it first and grant me the necessary rights there?

guybedford:
@asapach I think you need to initiate the transfer from within the original repo, then I can accept it here?

asapach:
https://help.github.com/articles/transferring-a-repository-owned-by-your-personal-account/
> Users must have admin or owner permissions within the receiving organization before they can transfer a repository that they individually own.

guybedford:
Ah right, added you now.

asapach:
Done. You should also have the NPM admin rights to the package.

guybedford:
Thanks, I'll just aim to run some tests locally here before merge but the approach seems 100%.

Just let me know if you have any other questions around this work.

guybedford:
I think the change is actually a valid fix - the `unknown` is what is reported by the transpiler source map so that is what should be used.

I've merged this in manually now.

It may also be worth abstracting a single `compiler.js` file which encapsulates the standard Babel transpilation up to the plugin parameter, as we're going to be repeating this code a lot otherwise.

PR #653, opened 2016-07-24
babel global transform

asapach:
I've changed the wrapper function slightly, because traceur wasn't handling directives properly:
```js
"use strict"
var foo = "bar";
```
Became this:
```js
(function() {
  var foo = this["foo"];
  "use strict";
  var foo = "bar";
  this["foo"] = foo;
})();
```
Now it becomes this:
```js
(function ($__global) {
  "use strict";
  var foo = $__global["foo"];
  var foo = "bar";
  $__global["foo"] = foo;
})(this);
```


guybedford:
Perhaps as discussed in the other issue we can generalize the Babel transform function now? Other than that I think we just need to rebase this to the latest master then it's good to merge?

I've added you back as an Admin to transfer the global transformer into the repo.

asapach:
I've created a new PR, because Github won't pick up my rebase: #654
Plugin transferred.

PR #680, opened 2016-08-30
Initial commit for amd transformer upgrade path

jrauschenbusch:
Hi @guybedford,

here is the promised PR. I've published the transformer plugin in an initial 0.0.1 version to npm.

Best regards

### DIFF
--- a/compilers/amd.js
+++ b/compilers/amd.js
@@ -1,4 +1,5 @@
-var System = require('systemjs');
+var compiler = require('./compiler');
+
 var traceur = require('traceur');
 var vm = require('vm');
 var traceurGet = require('../lib/utils').traceurGet;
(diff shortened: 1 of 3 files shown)


jrauschenbusch:
Won't be merged until all unit tests will pass.

@guybedford Can you help with investigating the problems? You can use my forked builder repo
https://github.com/jrauschenbusch/builder/tree/upgrade-amd-transformer.

jrauschenbusch:
It seems that the loading operation of several files fails as they are loaded from the wrong path:
```
test/output  **should be test/fixtures/test-tree**
  L jquery.js
    global.js
    plugins.js
    text_plugin.js
    first.js
    second.js
```

If the missing files are copied to test/output, then the following errors occur:

```
AMD 1, 3:
Error: (SystemJS) XHR error loading /builder/test/fixtures/test-tree/traceur
Unable to load transpiler to transpile /builder/test/output/first.js

AMD 4, 7:
Error: (SystemJS) Module /builder/test/output/amd-4.js interpreted as global module format, but called System.register.

AMD 11:
Error: (SystemJS) Multiple anonymous System.register calls in module /builder/test/output/amd-11.js. If loading a bundle, ensure all the System.register calls are named.

UMD:
Potentially unhandled rejection [16] Error: (SystemJS) Module /builder/test/output/umd.js interpreted as cjs module format, but called System.register.
```

PR #693, opened 2016-09-16
Add compiled outputs data for `bundle`, `buildStatic` methods

sergei-startsev:
I've added the additional field `outputs` to the output of `bundle`, `buildStatic` methods. The field contains a sorted array of sources and appropriate source map data before merging them to the single bundle:
outputs
It allows to add plugins that can build custom bundles based on provided data, e.g. `eval` bundle:
```javascript
eval(
   $__System.register(...);
   //# sourceMappingURL=moduleA.js.map
);
eval(...);
...
eval(...);
```
Here are more examples of different bundles that Webpack uses: [devtool](https://webpack.github.io/docs/configuration.html#devtool)

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -706,6 +706,7 @@ Builder.prototype.bundle = function(expressionOrTree, outFile, opts) {
         .then(function(output) {
           output.modules = compiled.modules;
           output.entryPoints = compiled.entryPoints;
+          output.outputs = compiled.outputs;
           output.tree = tree;
           output.assetList = compiled.assetList;
           if (outputOpts.outFile) {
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, I like the idea. How can the outputs object be used after being modified though? Can you provide an example of the sort of API you see this being used with?

sergei-startsev:
Here is how I've used it for building `eval` bundles (each module is executed with `eval` and a `sourceMappingURL` is also added to the eval):

```

**Comment by guybedford** [comment]
I think it would be better to accept an outputs hook as a build option something like:

```javascript
builder.bundle('x', {
  hook: {
    output: (outputs) => {
      return modifiedOutputs;
    }
  }
}).then(...);
```

That said, i'm weary to rush into a hook architecture on a whim without fully thinking it through as well.

lastmjs:
Could this pull request relate to my questions [here](https://github.com/systemjs/builder/issues/727) about a static build with the semantics of `compile`?

PR #694, opened 2016-09-18
fixed missing start-of-the-line column mappings

asapach:
I think I've finally tracked down the problems I was seeing with source maps thanks to @sergei-startsev. We've re-implemented the sourcemap concatenation routine using `SourceNode` and `SourceMapConsumer` from `source-map` instead of calculating offsets manually. @guybedford, if you're interested, it's here: https://github.com/asapach/builder/tree/sergei-startsev-sourcemaps

The major difference I've noticed is that current sourcemaps are missing mappings at the very start of the line (index 0). This can be reproduced in the following example:

```js
window.test = function () {
  return 'test';
};
```

Here are the source maps when this is bundled into a rollup bundle: [visualization link](http://sokra.github.io/source-map-visualization/#base64,KGZ1bmN0aW9uIChleHBvcnRzKSB7CiAgJ3VzZSBzdHJpY3QnOwoKICB3aW5kb3cudGVzdCA9IGZ1bmN0aW9uICgpIHsKICAgIHJldHVybiAndGVzdCc7CiAgfTsKCn0oKHRoaXMudGVzdCA9IHRoaXMudGVzdCB8fCB7fSkpKTsKLy8jIHNvdXJjZU1hcHBpbmdVUkw9YnVuZGxlLmpzLm1hcA==,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRlc3QuanMiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6Ijs7O1FBQU0sQ0FBQyxJQUFJLEdBQUcsWUFBWTtJQUN4QixPQUFPLE1BQU0sQ0FBQztHQUNmLENBQUMiLCJmaWxlIjoiYnVuZGxlLmpzIiwic291cmNlc0NvbnRlbnQiOlsid2luZG93LnRlc3QgPSBmdW5jdGlvbiAoKSB7XG4gIHJldHVybiAndGVzdCc7XG59O1xuIl19,d2luZG93LnRlc3QgPSBmdW5jdGlvbiAoKSB7CiAgcmV0dXJuICd0ZXN0JzsKfTsK)
And here's a plunkr repro: https://plnkr.co/edit/sc7IDhJYsZYFGTBOjsur?p=preview
Try setting a breakpoint on the first line in `test.js`:
breakpoint
And you'll see that even if you manage to do this, it will skip to line 3 instead:
breakpoint2

After this PR is applied the sourcemaps have the column mappings at the start of the line: [visualization link](http://sokra.github.io/source-map-visualization/#base64,KGZ1bmN0aW9uIChleHBvcnRzKSB7CiAgJ3VzZSBzdHJpY3QnOwoKICB3aW5kb3cudGVzdCA9IGZ1bmN0aW9uICgpIHsKICAgIHJldHVybiAndGVzdCc7CiAgfTsKCn0oKHRoaXMudGVzdCA9IHRoaXMudGVzdCB8fCB7fSkpKTsKLy8jIHNvdXJjZU1hcHBpbmdVUkw9YnVuZGxlLmpzLm1hcA==,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRlc3QuanMiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6Ijs7O0VBQUEsTUFBTSxDQUFDLElBQUksR0FBRyxZQUFZO0FBQzFCLEVBQUEsRUFBRSxPQUFPLE1BQU0sQ0FBQztBQUNoQixFQUFBLENBQUMsQ0FBQyIsImZpbGUiOiJidW5kbGUuanMiLCJzb3VyY2VzQ29udGVudCI6WyJ3aW5kb3cudGVzdCA9IGZ1bmN0aW9uICgpIHtcbiAgcmV0dXJuICd0ZXN0Jztcbn07XG4iXX0=,d2luZG93LnRlc3QgPSBmdW5jdGlvbiAoKSB7CiAgcmV0dXJuICd0ZXN0JzsKfTsK)

And here's the plunkr where the breakpoints are working properly: https://plnkr.co/edit/0StgiOwsRElAHBX9RjVY?p=preview

One minor issue is that there can be redundant mappings (related to whitespace?):
mappings
This is fixed in @sergei-startsev's branch mentioned above ([visualization link](http://sokra.github.io/source-map-visualization/#base64,KGZ1bmN0aW9uIChleHBvcnRzKSB7CiAgJ3VzZSBzdHJpY3QnOwoKICB3aW5kb3cudGVzdCA9IGZ1bmN0aW9uICgpIHsKICAgIHJldHVybiAndGVzdCc7CiAgfTsKCn0oKHRoaXMudGVzdCA9IHRoaXMudGVzdCB8fCB7fSkpKTsKLy8jIHNvdXJjZU1hcHBpbmdVUkw9YnVuZGxlLmpzLm1hcA==,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRlc3QuanMiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6Ijs7O0VBQUEsTUFBTSxDQUFDLElBQUksR0FBRyxZQUFZO0FBQzFCLElBQUUsT0FBTyxNQUFNLENBQUM7QUFDaEIsR0FBQyxDQUFDIiwiZmlsZSI6ImJ1bmRsZS5qcyIsInNvdXJjZXNDb250ZW50IjpbIndpbmRvdy50ZXN0ID0gZnVuY3Rpb24gKCkge1xuICByZXR1cm4gJ3Rlc3QnO1xufTtcbiJdfQ==,d2luZG93LnRlc3QgPSBmdW5jdGlvbiAoKSB7CiAgcmV0dXJuICd0ZXN0JzsKfTsK)), but it's more of a rewrite rather than a fix (or refactoring).


guybedford:
Amazing work, thanks so much for tracking this down!

The new concatenation work looks great as well. Would you be able to send a PR for this as well?

asapach:
As I've mentioned, it's a rewrite. There's a bunch of code in in `output.js` and `sourcemaps.js` that I don't quite understand, so I'm pretty sure there will be breaking changes and bugs. I think it would be worth the risk if it was the only way to fix the issue, but I'm happy with this smaller (ans safer) change for the time being. Unless you see other benefits to rewriting this piece of code.

guybedford:
@asapach a rewrite of those files is great and needed though! Is your version using less code or more code? If it's less that's already a win.

asapach:
OK, I'll see if I can refactor the code rather doing a big-bang rewrite.

guybedford:
👍  either is good, whatever you think is best.

PR #697, opened 2016-09-22
bumped rollup

asapach:
There are a lot of fixes since the last version: https://github.com/rollup/rollup/blob/master/CHANGELOG.md
I personally ran into this issue and would like to see it fixed: https://github.com/rollup/rollup/issues/668

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "es6-template-strings": "^2.0.0",
     "glob": "^7.0.3",
     "mkdirp": "^0.5.1",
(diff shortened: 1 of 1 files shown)


asapach:
Is builder still targeting Node 0.10? Because rollup evidently is not.

asapach:
0.10 is reaching end of support in October: https://github.com/nodejs/LTS#lts-schedule, so I think it should be safe to exclude it from travis config.

PR #704, opened 2016-10-01
prevent unnecessary recursion during trace

### LINKED ISSUE #701: Node mysteriously exits during .bundle
We are using the System.js builder API v0.15.32  to create a static bundle of our project (using both node 4.4.7 and 6.6.0). After recently changing some of our project's imports, we notice that the node process exits mysteriously when we are doing a `builder.bundle`. The promise never resolves or rejects and the process exits:

```
 builder.bundle('ourproject/**/*.js', 'bundles/app.js', {
  browser: true,
  production: true,
  minify: true,
  mangle: false,
  sourceMaps: false
}).then(function(output) {
  // this is never reached
}, function(err) {
  // nor is this
});
// ...but the process exits
```

Through a series of `console.log` statements, the offending code has been determined to be contained within `traceExpression` inside `arithmetic.js`.

Unfortunately, our project is quite large and because of the lack of debug output from system.js and the recursive nature of imports, we cannot find a small reproduction of this. In our code, in select files, we can comment out groups of `import` statements and the build will then start succeeding:

```
...
// as long as these are *all* commented the above issue does not happen
// import from 'a.js'
// import from 'b.js'
// import from 'c.js'
```

Any advice on troubleshooting this? There is zero output and `process.on('uncaughtException')` is never triggered. Our temporary solution is going to be selectively loading some modules asynchronously at runtime, but the fact that is happens based on the structure of our imports is concerning.

ghempton:
Fixes #701 

The current exit condition is sub-optimal. As the tree of load dependencies is traversed asynchronously, it is possible that the same module's dependencies could be recursed many times. In our case, this was triggering a massive number of promises and crashing node itself.

### DIFF
--- a/lib/trace.js
+++ b/lib/trace.js
@@ -628,6 +628,11 @@ Trace.prototype.getAllLoadRecords = function(canonical, traceOpts, canonicalCond
   var self = this;
   return this.getLoadRecord(canonical, traceOpts, parentStack)
   .then(function(load) {
+
+    // check this again – might have been loaded asynchronously since the last check
+    if (canonical in curLoads)
+      return curLoads;
+
     // conditionals, build: false and system modules are falsy loads in the trace trees
     // (that is, part of depcache, but not built)
     // we skip system modules though


ghempton:
FWIW, this also gives us a *massive* performance boost locally:

Before:

```
./bin/build  146.57s user 6.64s system 101% cpu 2:31.69 total
```

After:

```
./bin/build  54.98s user 4.35s system 101% cpu 58.374 total
```

guybedford:
getLoadRecord itself does have a synchronous queue through the `tracing` object at https://github.com/systemjs/builder/blob/master/lib/trace.js#L140.

But this will avoid duplicate tracing work on the tree certainly, which would be especially important with deep module trees.

Thanks so much for tracking this down!

ghempton:
Mind releasing a new version?

PR #708, opened 2016-10-04
Allow specifying mangle options

ghempton:
Uglifyjs supports options passed into the `mangle_names` method: https://github.com/mishoo/UglifyJS2/blob/4761d07e0bc3d4c53e0c9c72fc9c322c95cb090e/lib/scope.js#L394

This PR changes the `mangle` option to optionally take an object and pass options.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -79,11 +79,15 @@ function minify(output, fileName, mangle, uglifyOpts) {
     throw new Error(e);
   }
   ast.figure_out_scope();
-  
+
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, it might be nice if we rather retain the existing functionality, but just add an extra `uglifyOpts.mangle` option? We support `builder.build('x', { uglify: { compress: {...custom compressor options..}, beautify: {...} } }` so could just extend that to `{ uglify: { compress: {...}, mangle: {...}, ... } }`.

PR #718, opened 2016-10-18
pass "compact" option to esm compiler based on systemjs configuration

sormy:
Pass "compact" option to esm compiler based on systemjs configuration.
This allows to create non minified development builds for applications larger than 100kb.
Code is easy to fix to pass other options besides "compact"


guybedford:
Thanks, I've added this manually in https://github.com/systemjs/builder/commit/d20a7f772fa8c44f257cc925ab217488fd25d336. The `loader.babelOptions` are not supposed to apply in general at this layer.

Issue #264, opened 2015-08-04
Full System.register AMD wrapping


guybedford:
It would be possible to make AMD a full wrapped format when bundled like the others as `System.register` by applying transforms similar to the global transforms we have for handling globally-scoped assignments.

A fully-wrapped `System.register` output would simplify the meaning of an AMD bundle and unify the ability to subtract bundles, which currently needs some exceptions to work.

Rush:
Maybe it's related ... update from systemjs/builder from 0.13.x to 0.14.5 broke import of AMD modules using `System.import("application.js");` .. I had to load it from `System.amdRequire(["application.js"])`. I understand that if this issue was fixed, I could again use `System.import`.

guybedford:
@Rush that sounds like something else. `System.amdRequire` just calls `System.import` so I'm not sure how that would have fixed it.

guybedford:
Full explanation:

Right now, AMD is replaced in-place:

```javascript
function global() {
}

var anotherGlobal = 5;

// yes we allow named defines alongside the define below
define('x', function() {
});

if (thing)
  define(function() {
    return { a: global, b: anotherGlobal };
  });
```

Where we just replace the `define` statement itself with `System.registerDynamic('...', ...)` to ensure the exact same execution behaviour. The existing named define gets left as-is.

The issue with this is that AMD is then not a "wrapped" format... it still runs with its expected bindings. So we're not doing strict conversion like we do with the others where we have a reliable form of:

```javascript
System.registerDynamic(..., function() {
  // all code is wrapped in here
});
```

In order to turn AMD into a wrapped format, this involves doing the same thing we do for global transformers by assigning variables to the bindings:

```javascript
System.registerDynamic('y', ..., function(require, exports, module) {
  // a standard prefix?
  var define_value;

  function global() {
  }

  var anotherGlobal = 5;

  // named define gets converted into a nested System.registerDynamic!?
  System.registerDynamic('x', [], false, function() {
    // ...
  });

  if (thing)
    // we convert the define into an IIFE
    define_value = (function() {
      // exports and require in scope need to be handled
      return { a: global, b: anotherGlobal };
    })();

  // update global assignments back to global
  this.global = global;
  this.anotherGlobal = anotherGlobal;

  // return
  return define_value;
});
```

The other benefit of this would be that we no longer need to rely on the global `define` being present as well.

This is quite complex because AMD has so many frickin forms... but I think it would be a nice opportunity to start.

guybedford:
[...]

PR #722, opened 2016-10-19
Upgrade amd transformer to babel

guybedford:
PR for https://github.com/systemjs/builder/issues/264.

//cc @jrauschenbusch


lib/output.js
```diff
@@ -46,7 +46,6 @@ function processOutputs(outputs) {
     else {
       throw "Unexpected output format: " + output.toString();
     }
-    source = removeSourceMaps(source || '');
```
guybedford: Reason for removal?

lib/output.js
```diff
@@ -148,12 +147,11 @@ exports.writeOutputs = function(outputs, baseURL, outputOpts) {
   var fileName = outFile && path.basename(outFile) || 'output.js';
 
   var output = createOutput(outFile || path.resolve(basePath, fileName), outputs, basePath, outputOpts.sourceMaps, outputOpts.sourceMapContents);
-
+	
   if (outputOpts.minify)
     output = minify(output, fileName, outputOpts.mangle, outputOpts.uglify);
 
   if (outputOpts.sourceMaps == 'inline') {
-    output.source = inlineSourceMap(output.source, output.sourceMap);
```
guybedford: And here?

templates/sfx-core.js
```diff
@@ -274,7 +274,7 @@
       throw new TypeError('Module ' + name + ' not declared as a dependency.');
     }, exports, module);
 
-    if (typeof output !== undefined)
+    if (output)
```
guybedford: And again?

test/sourcemaps.js
```diff
@@ -31,14 +31,12 @@ var readExpectation = function(filename) {
...
 
-function writeTestOutput(done) {
+function writeTestOutput() {
   var builder = new Builder();
   return builder.loadConfig(configFile)
     .then(function() {
       builder.buildStatic('first.js', 'test/output/output.js', buildOpts);
-    })
-    .then(done)
-    .catch(err);
```
guybedford: ?

test/test-build.html
```diff
@@ -145,7 +148,7 @@
           return System['import']('amd-6a.js');
         })
         .then(function(m) {
-          ok(window.p == 'a');
+          ok(m.p == 'a');
```
guybedford: Why not the `window` check?

compilers/compiler.js
```diff
@@ -31,4 +34,16 @@ exports.compile = function (load, opts, plugin) {
...
+  var options = {
+    plugins: [[require('babel-plugin-transform-amd-system-dependencies').default]]
+  };
+
+  var output = babel.transform(load.source, options);
+
+  return {
+    amdDeps: output.metadata.amdDeps
+  };
+}
```
guybedford: This should probably still use the default Babel options.

Also we should add `ast: true` so we can return and cache the ast and not need to parse twice.

Then we can probably just return the Babel output directly and call this a more general `exports.compileAst` which returns `metadata` and `ast`.

compilers/amd.js
```diff
@@ -360,34 +263,17 @@ exports.remap = function(source, map, fileName) {
...
-    sourceMapOffset: 2
-  });
+exports.compile = function (load, opts, loader) {
+  opts.moduleId = !opts.anonymous && load.name; 
+  return compiler.compile(load, opts, [require('babel-plugin-transform-amd-system-wrapper').default, {
+    map: function (dep) {
+      return opts.normalize ? load.depMap[dep] : dep;
+    },
+    systemGlobal: opts.systemGlobal,
+	deps: load.deps
```
guybedford: Are these deps additive to the existing deps?

guybedford:
As mentioned previously I'm still not sure why we need two separate projects to do the same thing here. A single project could expose both variants of the transform I think through options.

Also ast caching would be super useful - we then only parse once in `attach` and can store the ast on `metadata` for the load record, optionally picking it up in the compile method. We do this already for `exports.attach` in `compilers/esm.js` as an example.

guybedford:
This all seems great to me otherwise!

guybedford:
And @jrauschenbusch I should say again that this is some really amazing and critical work I never would have had the time to do. Thank you so so much for your efforts.

test/test-build.html
jrauschenbusch: It seems that the new wrapping format adds the property p just to the module exports and no longer to the window namespace.

```javascript
"bundle";
System.registerDynamic('amd-6a.js', [], false, function ($__require, $__exports, $__module) {
  return (function (require) {
    this.p = 'a';
  }).call($__exports, $__require);
});
```

compilers/compiler.js
jrauschenbusch: Sounds very good! This could be a real performance gain! The function name `transformAMDDependencies` already had the appearance not to belong in the compiler part.

compilers/amd.js
jrauschenbusch: Initially i didn't used this dependencies. But some unit tests still failed, and i had no idea what's wrong. Then i looked into the original code and found this code snippet: https://github.com/systemjs/builder/blob/master/compilers/amd.js#L222

It sems there are additional deps only available in the loader metadata.

lib/output.js
jrauschenbusch: This tries to inline the sourcemap, but output.sourceMap is empty when using babel, since the source map is already present in output.source (if `removeSourceMaps(source || '')` isn't called before.

lib/output.js
jrauschenbusch: The reasons is that babel already includes the sourcemaps into the output.source. output.sourceMap is emtpy, hence this just removes the already existing sourcemap.

jrauschenbusch:
@guybedford It's quite clear that no single person is able to do each proposed/desired optimization in such an evolved ecosystem like SystemJS/SystemJS Builder and JSPM. It's a pleasure for me to help you and it is cool to be able to bring this ecosystem to the next level.

Regarding your ideas:

> As mentioned previously I'm still not sure why we need two separate projects to do the same thing here. A single project could expose both variants of the transform I think through options.

Absolutely true. There is no need for two separate projects. We should get the complete logic into one single babel transformer which can be parameterized.

> Also ast caching would be super useful - we then only parse once in attach and can store the ast on metadata for the load record, optionally picking it up in the compile method. We do this already for exports.attach in compilers/esm.js as an example.

Sounds really good. Would be great to have some support here.

guybedford:
Thanks @jrauschenbusch I can do the work to get the shared ast, if you're able to look at refactoring the plugins into one that would be great. Otherwise feel free to transfer the repo across into this org and I can continue to try and work on it when I eventually have time.

guybedford:
I would really love to get this merged, along with the latest CJS transformer updates. It makes the transformations fully independent of the SystemJS builder applying more generally to https://github.com/ModuleLoader/es-module-loader as well.

@jrauschenbusch what is the current status? How can I help to move this forward? (I'm happy to dive into code / contribute feedback / financially etc)

jrauschenbusch:
Hi @guybedford,

I have not worked on it in the last few weeks because I am fully strechted with my primary work until Christmas. But I will use the contemplative christmas time to merge the AMD transformer into a single plugin. We could discuss more about this process in the next days. Maybe we can also work together on it to get it ready for a first integration.

As you maybe have recognized i've transfered the repos the the systemjs org. :-)

guybedford:
@jrauschenbusch thanks for the continued updates here despite the holidays! There seems to just be that one source maps test failing here in the CI. Do you have any idea what this is about?

jrauschenbusch:
Hi @guybedford. This PR is not finished yet for a full merge. I just want to rewrite some parts of the builder itself (cjs/amd remap functions and the shared ast part related to the compiler.js (exports.compileAst) part.

But i've already merged the logic of https://github.com/systemjs/babel-plugin-transform-amd-system-dependencies into https://github.com/systemjs/babel-plugin-transform-amd-system-dependencies. Hence the first repository will not be needed anymore, as it is also not published via npm.

I think i will finish my work until next week.

test/sourcemaps.js
jrauschenbusch: I do not remember why I've deleted the `done` callback trigger. Maybe `done` was undefined.

jrauschenbusch:
Hi @guybedford @asapach,

Now we have a first working solution.

Some notes :
- I've removed the remap functions, as it seems they aren't referenced anymore in the project.
- The shared AST optimization doesn't work inside the test environment, since https://github.com/systemjs/builder/blob/master/lib/trace.js#L596 resets the stored ast to undefined. If i comment out this line an ESM test fails.

Please perform a code review of both, this PR and the last release of https://github.com/systemjs/babel-plugin-transform-amd-system-wrapper.

What we should consider to test:
- Building of large applications
- Source maps (quality)
- Source comments (presence)
- Performance

asapach:
@jrauschenbusch, I've done some sanity checks and it seems like `{ sourceMaps: 'inline' }` option doesn't work: instead of one `sourceMappingURL` comment at the end, they are scattered after each individual file in the bundle. It looks like when we pass `opts.sourceMaps` to babel we need to convert them to bool: `!!opts.sourceMaps`.

asapach:
Added a fix: #753. It now passes on master, but fails after this PR because of this line: https://github.com/systemjs/builder/pull/722/files#diff-00fe37b20ddfb9ecc9c492b0725fa2b2L157

lib/output.js
```diff
@@ -47,7 +47,6 @@ function processOutputs(outputs) {
       var format = output ? output.toString() : typeof output;
       throw new Error("Unexpected output format: " + format);
     }
-    source = removeSourceMaps(source || '');
```
asapach: don't remove this line

lib/output.js
asapach: keep this line

guybedford:
Thanks so much @asapach for helping with review here. I'll check this when you're happy if that's ok?

lib/output.js
jrauschenbusch: Fixed

lib/output.js
jrauschenbusch: Fixed

jrauschenbusch:
@asapach I've implemented the requested changes.

@guybedford It seems that with the update to SystemJS 0.19.42 the Promise variable cannot be found anymore. Hence the TravisCI build are failing.

```bash
1) Test Tree Bundle Loading AMD from a bundle:
     ReferenceError: Can't find variable: Promise
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/systemjs/dist/system.src.js:840
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/systemjs/dist/system.src.js:1811
      at file:///media/jochen/DATA/systemjs/builder-2/test/test-build.html:37
      at callFnAsync (file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4506)
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4446
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4934
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:5071
      at next (file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4854)
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4864
      at next (file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4796)
      at file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:4832
      at timeslice (file:///media/jochen/DATA/systemjs/builder-2/node_modules/mocha/mocha.js:75)
```

compilers/compiler.js
```diff
@@ -20,15 +20,33 @@ exports.compile = function (load, opts, plugin) {
...
-};
\ No newline at end of file
+};
+
+exports.compileAst = function (load, plugin) {
+  return babel.transform(load.source, {
+      babelrc: false,
+      compact: false,
+      filename: load.path,
+      inputSourceMap: load.metadata.sourceMap,
```
asapach: if you're consuming the source map, maybe you should produce one as well? i.e. like `sourceMaps` option above

test/test-build.html
```diff
@@ -17,14 +17,17 @@
...
+      },
       paths: {
         'traceur-runtime': '../node_modules/traceur/bin/traceur-runtime.js',
         'babel/external-helpers': '../node_modules/babel-core/external-helpers.js',
         'jquery-cdn': './jquery.js',
         'output/*': './output/*'
       },
       bundles: {
-        'output/tree-build.js': ['third.js', 'cjs.js', 'jquery.js', 'second.js', 'global.js', 'amd.js', 'first.js'],
+        'output/tree-build.js': ['first.js', 'second.js', 'third.js', 'cjs.js', 'jquery.js', 'global.js', 'amd.js'],
```
guybedford: Changing the dependency order in the build is a breaking change. We need to ensure that dependencies are reported in the same order.

test/test-build.html
```diff
@@ -17,14 +17,17 @@
 
     System.config({
       baseURL: './fixtures/test-tree/',
+      meta: {
+         'output/*': { format: 'register' }
+      },
```
guybedford: Is this really necessary? The detection should be maintained here surely?

guybedford: I think for this reason we may need to make this change a 0.16 builder PR, instead of adding this to 0.15.

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "babel-core": "^6.9.0",
+    "babel-plugin-transform-amd-system-wrapper": "^0.1.0",
```
guybedford: Does this mean we can finally remove `traceur` as a dependency as well!?

guybedford:
I've rebased the PR to SystemJS 0.16, but there are some conflicts that need to be resolved for this to work.

test/test-build.html
jrauschenbusch: :+1:

package.json
jrauschenbusch: There are still some references to traceur in /compilers/esm.js and in some /lib/*.js files. We have to remove them first before we can remove traceur completely.

test/test-build.html
jrauschenbusch: Ok, so yo mean we should revert the order?

guybedford: Yes please, it is important to ensure the same ordering of dependencies (which should be a post order traversal of the tree ordered by require order). We didn't discuss this before right?

package.json
guybedford: Ok, that seems all the uses are then just the `getDeferredImports` function (which takes a source string, and returns the arguments of all `System.import('x')` calls in that source.), and the ESM remap function (which takes an esm source, and returns a new esm source with the imports remapped by the given map function).

Both of these are completely separate things though so can be a separate PR at some point in future. There is no rush to upgrade these.

I think just getting this transformer ready for 0.16 would be ideal fixing up the orderings.

test/test-build.html
jrauschenbusch: Ok, done. No, we didn't have discussed this before.

package.json
jrauschenbusch: :+1: Hint: I've removed the remap function in the cjs and amd tranfsformers as they aren't referenced by any code.

jrauschenbusch:
@guybedford I've resolved the conflicts. But now the build breaks again because of :

```bash
  1) Source Maps can handle multiple transpiled files with same name:
     Uncaught Error: The globalName option must be set for full-tree rollup global and UMD builds.
```

Do you have any clue why this is so?

guybedford:
@jrauschenbusch I think that is the failing build on 16 actually, which I'm working on right now.

Can you just rebase against 0.16 in the mean time as well (git pull --rebase origin 0.16) so that the conflicts in this PR are resolved?

guybedford:
(sorry I just updated 0.16 again!)

guybedford:
@jrauschenbusch I think we're good to merge. Can you perhaps try a `git pull --rebase` once more as it seems the merge ended up creating some duplicate commits in https://github.com/systemjs/builder/pull/722/commits. These may dedupe with a rebase, which would be cleaner to land.

guybedford:
Moved to https://github.com/systemjs/builder/pull/764.

Issue #734, opened 2016-11-11
Webpack dies when requiring a UMD build created by systemjs-builder


joeldenning:
Projects built into a UMD format by systemjs-builder will cause build-time warnings and runtime exceptions when included in a webpack bundle. The reason is the [dynamic usage of `require` inside sfx-core.js](https://github.com/systemjs/builder/blob/master/templates/sfx-core.js#L370), which [webpack detects](https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/lib/dependencies/CommonJsRequireDependencyParserPlugin.js#L29) at build-time. Also, for reference, [here](https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/lib/dependencies/WebpackMissingModule.js#L22) is where webpack injects code that throws an error at runtime under this scenario.

https://github.com/CanopyTax/single-spa/issues/79 is one project being affected by this.

Depending on how you look at it, this problem could be blamed on webpack or on systemjs-builder. The case for it being webpack's fault is that systemjs-builder is [not even calling the require function dynamically](https://github.com/systemjs/builder/blob/master/templates/sfx-core.js#L370) under the UMD scenario, since webpack treats UMD modules as AMD modules and AMD modules don't call `require`. On the other hand, the case for it being systemjs-builder's fault is that there isn't a way for webpack to statically analyze the code and know that the `nodeRequire` isn't going to be called later on.

I'm interested to hear your perspective on this, and plan on putting together a pull request that is a best effort at solving the problem inside of systemjs-builder. Additionally, I'd love to hear an explanation of why System._nodeRequire is being stored even inside of UMD builds.

PR #735, opened 2016-11-11
Fixing #734. Not using 'require' if process.browser is set to true

joeldenning:
See https://github.com/systemjs/builder/issues/734. In that issue, there are two related problems described: (1) webpack warnings, and (2) runtime error. This pull request will reduce the number of webpack warnings from two to one, and it will eliminate the runtime error. Afaik, getting rid of that last webpack warning would only be possible if we do not store System._nodeRequire, which does not seem like a likely option.

Let me know if there's something better to do here to fix the problem; I'm only about 50%-75% sure that this is a good way to resolve the issue.

Also note that this would fix https://github.com/CanopyTax/single-spa/issues/79

### DIFF
--- a/templates/sfx-core.js
+++ b/templates/sfx-core.js
@@ -367,7 +367,7 @@
     entry.module.execute.call(global);
   }
 
-  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof process != 'undefined' && !process.browser && typeof require != 'undefined' && typeof require.resolve != 'undefined' && require;
+  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof require != 'undefined' && typeof require.resolve != 'undefined' && typeof process != 'undefined' && process.platform && require;
 
   // magical execution function
   var modules = { '@empty': {} };
--- a/templates/sfx-core.js
+++ b/templates/sfx-core.js
@@ -367,7 +367,7 @@
     entry.module.execute.call(global);
   }
 
-  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof require != 'undefined' && require.resolve && typeof process != 'undefined' && require;
+  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof process != 'undefined' && !process.browser && typeof require != 'undefined' && typeof require.resolve != 'undefined' && require;
 
   // magical execution function
   var modules = { '@empty': {} };


templates/sfx-core.js
```diff
@@ -367,7 +367,7 @@
     entry.module.execute.call(global);
   }
 
-  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof require != 'undefined' && require.resolve && typeof process != 'undefined' && require;
+  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof process != 'undefined' && !process.browser && typeof require != 'undefined' && typeof require.resolve != 'undefined' && require;
```
joeldenning: Here are the things I changed:

1) We now only set nodeRequire to be the `require` variable when `!process.browser`. As far as I can tell, process.browser seems to be a browserify thing that webpack and other build systems have also adopted. So I'm hoping it is the right thing to check, since we do not want to set `nodeRequire` to be the nodejs `require` if our code is being bundled for the browser. I feel pretty good about that for the UMD case, but am not as sure for the other SFX use cases.
2) I moved the `typeof process` check to the beginning, so that the process.browser check can be near the beginning so that we short circuit out of the conditional expression before hitting the `&& require` that webpack replaces with a `throw` statement. This is what fixes the runtime error that is thrown by webpack (webpack replaces `&& require` with a `throw` statement.
3) I changed `&& require.resolve` to `&& typeof require.resolve != 'undefined'` because webpack knows that a typeof statement doesn't count as a dynamic usage of `require.resolve`. This is what eliminates one of the webpack build-time warnings.

guybedford:
Thanks for the detailed explanation here.

I think we do still need to set `nodeRequire` in the `process.browser` case though in order to support Electron unfortunately.

Is there an adjustment we can make to this to ensure it is still set?

joeldenning:
Here are some other options that each would fix the problem:

1) check for process.platform, which is undefined in webpack
```
var nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof require != 'undefined' && typeof require.resolve === 'function' && typeof process != 'undefined' && process.platform && require;
```
2) try/catch the whole thing so that if webpack throws an error at runtime, nothing dies:
```
var nodeRequire;
try {
  nodeRequire = typeof System != 'undefined' && System._nodeRequire || typeof require != 'undefined' && typeof require.resolve === 'function' && typeof process != 'undefined' && require;
} catch(err) {
  nodeRequire = undefined;
}
```
3) A more explicit try/catch
```
var canUseRequire;
try {
  canUseRequire = typeof require !== 'undefined' && typeof require.resolve === 'function' && !!require;
} catch(err) {
  canUseRequire = false;
}
var nodeRequire = typeof System != 'undefined' && System._nodeRequire || canUseRequire && typeof process != 'undefined' && require;
```
4) Check for webpack directly:
```
  var nodeRequire = typeof System != 'undefined' && System._nodeRequire || !__webpack_require__ && typeof require != 'undefined' && typeof require.resolve === 'function' && typeof process != 'undefined' && require;
```

Which do you think is best?

guybedford:
The `process.platform` approach sounds good to me.

joeldenning:
Sounds good - I've updated the pr.

PR #750, opened 2017-01-02
added test case "can handle multiple transpiled files with same name"

arty-name:
Here is a test case to illustrate an issue I have found: under certain conditions Builder throws an error `Uncaught Error: Mismatched sourcesContent for: amd.js`. 

The background: I have some AMD code which already uses elements of ES6. I try to gradually migrate it to Builder. I use `plugin-babel` do to the transpilation, and I need source maps. However with these requirements two files with identical names cause the Builder to hiccup.


asapach:
I've stepped through this, and I think the issue is in `plugin-babel` and this PR appears to fix it: systemjs/plugin-babel#71
@guybedford, could you please publish a new release?

guybedford:
@asapach sure I've released plugin-babel 0.0.18. Let me know if that works out ok here.

asapach:
Yup, now tests pass for me. @arty-name try bumping to 0.0.18 to see if it fixes your issue.

arty-name:
Thank you, @guybedford and @asapach, that has resolved the issue. 

Would you like to merge this PR to avoid regressions in the future?

guybedford:
@arty-name sure extra tests can never hurt :) Thanks for your efforts here.

arty-name:
Seems like builds on Node 0.x have failed because of missing module `process`. Does plugin-babel support these versions?

guybedford:
That's just the plugin support yeah. But 0.12 is EOL anyway so it's not a big deal.

PR #753, opened 2017-01-14
fixed inline source maps

asapach:
Fixes `{ sourceMaps: 'inline' }` case. Babel was unnecessarily adding inline source maps to the output - instead it should add them to the metadata.


guybedford:
Thanks.

PR #756, opened 2017-01-17
Fix { sourceMaps: 'inline' } option; fixes #754

### LINKED ISSUE #754: sourceMaps: 'inline' producing empty listing (test case included)
@guybedford For some reason providing `sourceMaps: 'inline'` to the builder produces a sourcemap listing containing nothing.

Setting `sourceMaps: true` gives me a full set of sourceMaps. But `sourceMaps: 'inline'`, via an in-memory build, produces 
```js
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IiIsImZpbGUiOiJvdXRwdXQuanMiLCJzb3VyY2VzQ29udGVudCI6W119
```

Decoding the base64 gives: 
```json
{"version":3,"sources":[],"names":[],"mappings":"","file":"output.js","sourcesContent":[]}
```

To reproduce:
```
git clone https://github.com/alexisvincent/systemjs-tools
cd systemjs-tools/test
./test-sourcemaps.sh
```

This will download the npm and jspm deps and then run the build server. To see the bundled file produced, navigate to `https://localhost:3000/dependencies.js`.

To play with the builder options, you can modify the `test/package.json` `systemjs-tools.builder.options` key. 

Note, the options specified there are merged in with the defaults specified [here](https://github.com/alexisvincent/systemjs-tools/blob/master/lib/config.js#L104), in otherwords, these:
```js
{
  sourceMaps: 'inline',
  production: false
}
```

If there's anything more I can do, please let me know.

Note: `systemjs-tools` uses a cross session builder cache. So to get a true fresh start, delete the generated `test/.systemjs.cache.json`

alexisvincent:
Resolves issue where only source maps of `jspm_packages` were being included when building `{ sourceMaps: 'inline' }` option.

### DIFF
--- a/compilers/esm.js
+++ b/compilers/esm.js
@@ -94,7 +94,7 @@ exports.compile = function(load, opts, loader) {
       plugins: [[require('babel-plugin-transform-es2015-modules-systemjs'), { systemGlobal: opts.systemGlobal }]],
       filename: load.path,
       //sourceFileName: load.path,
-      sourceMaps: opts.sourceMaps,
+      sourceMaps: !!opts.sourceMaps,
       inputSourceMap: load.metadata.sourceMap,
       moduleIds: !opts.anonymous,
       moduleId: !opts.anonymous && load.name,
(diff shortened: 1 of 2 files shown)


guybedford:
:+1:

alexisvincent:
Awesome :) eta on release?

Issue #779, opened 2017-03-07
trace uses 'system' for the format while rollup and compile checks for 'register'


zenorbi:
I'm trying to `buildStatic` a module to `format: "cjs"` and that module has `System.register` format. The builder's trace correctly identifies the module as a `register` format (https://github.com/systemjs/builder/blob/8200f4973faf93d079441cf8edbced3c260c8cc4/lib/trace.js#L576) but it turns it into `system` (presumably for consistency). However other parts of the builder (https://github.com/systemjs/builder/blob/8200f4973faf93d079441cf8edbced3c260c8cc4/lib/compile.js#L466) and (https://github.com/systemjs/builder/blob/37b73424e25e1b2871847108f333bca0431a5c34/lib/rollup.js#L257) uses the `register` name, thus not correctly detecting if the output should use the default export for the output or the whole module.

The result is that if I try to `buildStatic` this:

```js
System.register([], function (exports_1, context_1) {
    "use strict";
    var __moduleName = context_1 && context_1.id;
    var hello;
    return {
        setters: [],
        execute: function () {
            hello = "world";
            exports_1("hello", hello);
        }
    };
});
```

and require it using `require("./built-file")` the result is `undefined`.

zenorbi:
I'm willing send in a PR but I need to know which format should be the chosen one, `system` or `register`.

guybedford:
@zenorbi a PR here would be great. The preferred name is `system` at this point.

PR #781, opened 2017-03-09
Fix for issue #779

zenorbi:
This PR fixes https://github.com/systemjs/builder/issues/779. I had to upgrade phantomjs from `1.9.18` to `2.1.7` because `1.9.18` always exited with code 0 thus not really testing. Also I couldn't provide a good test for `rollup.js` so I just added checks for both `system` and `register` format.


guybedford:
Awesome, thanks for upgrading the phantom tests. I think it should be ok with `system` in the rollup code? Perhaps just add a throw / assertion at that point if the format is register so we can catch the bug if it comes up?

guybedford:
Thanks for your work here :)

PR #812, opened 2017-06-12
buildStatic support for context without a global

chrmod:
I'm building a sfx bundle to be run inside of Firefox process script. 

In that context, explicit global object (ContentProcessMessageManager) does not happen to accessible via variable. 

This PR aims to generate a bundle that will not throw errors on missing global in process script context.

Not sure if using this instead of self or global makes sense, but at least no errors are thrown and bundle works as expected. Please help to improve if there is a way.

Some mozilla docs:
* https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIProcessScriptLoader
* https://developer.mozilla.org/en-US/docs/User:wbamberg/Multiprocess_Firefox-2/Message_manager/Message_manager_classes#ContentProcessMessageManager


chrmod:
it does not really work atm, I'm going to update the PR as soon as it will also making the tests green.

chrmod:
the problem was wrong version of uglify-js used for minify.sh

guybedford:
Thanks, I'll post this out with the coming release cycle.

PR #815, opened 2017-06-28
Uglify Upgrade

timfish:
This is an initial stab but needs more work. 

Now I'm looking through the code I realise the options merging/transformations should really be happening in `processOutputOpts` but there is quite a bit of legacy stuff in there without tests so I'll need to be careful not to break anything.

It should support the existing config options:
```
builder.bundle('myModule.js', 'outfile.js', { minify: true, mangle: false, uglify: uglifyOps  });
```
Or you can pass the entire uglify-es valid config in yourself through `options.uglifyes`:
```
var uglifyOpt = { 
    compress: {
        global_defs: { DEBUG: false },
        dead_code: true,
    },
    mangle: false,
    warnings: false 
}
builder.bundle('myModule.js', 'outfile.js', { minify: true, uglifyes: uglifyOpt });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "source-map": "^0.5.3",
     "systemjs": "^0.19.43",
     "traceur": "0.0.105",
-    "uglify-js": "~2.7.5"
+    "uglify-es": "https://github.com/mishoo/UglifyJS2#harmony"
   },
   "devDependencies": {
     "babel": "^5.8.38",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "systemjs-builder",
-  "version": "0.15.35",
+  "version": "0.15.36",
   "description": "SystemJS Build Tool",
   "main": "index.js",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "source-map": "^0.5.3",
     "systemjs": "^0.19.43",
     "traceur": "0.0.105",
-    "uglify-js": "^2.6.1"
+    "uglify-js": "~2.7.5"
   },
   "devDependencies": {
     "babel": "^5.8.38",
--- a/lib/output.js
+++ b/lib/output.js
@@ -82,7 +82,7 @@ function minify(output, fileName, options) {
       compress: options.uglify.compress,
       mangle: options.mangle,
       output: {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "systemjs-builder",
-  "version": "0.15.35",
+  "version": "0.15.36",
```
guybedford: This seems to be a PR against the 0.16 branch instead of the 0.15 one? Happy to work with either though.

timfish: 0.16 will be fine. I'm happy to help them get aurelia-bundler working with 0.16 and then 0.15 support wont matter to me!

guybedford:
@timfish see also the approach in https://github.com/systemjs/builder/pull/818. I wonder how far a 3.x upgrade gets us here?

timfish:
The v3/master readme says:

```
- uglify-js@3 has a simplified API and CLI that is not backwards compatible with uglify-js@2.
- uglify-js only supports ECMAScript 5 (ES5).
- Those wishing to minify ES2015+ (ES6+) should use the npm package uglify-es.
```

I can't see any obvious features that you'd get updating to v3/master. The api changes to v3 are not backwards compatible so I doubt you can just upgrade the package [like that](https://github.com/systemjs/builder/pull/818/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L23).

guybedford:
Ok thanks for checking this. We should definitely stick with this upgrade path then.

behzad888:
I want to using `babel-plugin-syntax-dynamic-import` in bundler task, but i have an error in this case. have anybody an idea?
for more details you  can see > https://travis-ci.org/behzad888/builder/builds/253923865

guybedford:
@behzad888 it looks like you're running an older version of SystemJS builder, try upgrading to the latest.

behzad888:
@guybedford i'm using latest version of SystemJS builder but new version of uglify-js need to define transpiler `SystemJS.config({ transpiler: 'transpiler-module' })`. Although in my local project its ok but travis ci builder has an error as `Unable to dynamically transpile ES module`.  
How to set transpiler on SystemJS builder test cases?

timfish:
Seems like uglify-es is no longer supported.

guybedford:
See ‘terser’, it’s replacement fork.

Issue #814, opened 2017-06-25
relative dependencies are resolved from project root when builder is given windows style paths


CxRes:
Let there be a file, say:
```
./foo/bar/index.js
```
that invokes
```
import baz from './baz'
```

Now this works:
```
jspm bundle ./foo/bar/index.js myBundle.js
```

But this throws an error
```
jspm bundle .\foo\bar\index.js myBundle.js

err  Error on fetch for baz.js at file:///D:/path/to/project/baz.js
        Loading foo\bar\index.js
        Error: ENOENT: no such file or directory, open 'D:\path\to\project\baz.js'
```

I think this is a regression as this worked well with jspm 0.16.31 which used builder 0.15.11. I an guessing that the issue arose after 0.15.14 builder update.
node 8.1.0
npm  5.0.3
jspm 0.16.53

guybedford:
Try using `jspm bundle ./foo/bar/index.js myBundle.js` here. Module names exist in URL space, and the bundle argument is a module name so should be `/`-separated. `myBundle.js` is a file name so that is ok to be `\`-separated though.

CxRes:
@guybedford I figure that and have already fixed my code accordingly (though it goes against expectation for what an input "path" is). 

But this MUST be clearly documented, perhaps with an `important!` preceding it because it can be very easily missed and cause much pain for something that is rather trivial.

Will leave this open until the patch is accepted

PR #816, opened 2017-06-28
Added warning that module names exist in URL space

CxRes:
Added a warning that module names exist in URL space, and the bundle, buildStatic and trace argument is a module name so should be `/` separated. In reference to #814.


guybedford:
Sure, thanks!

PR #818, opened 2017-07-08
update uglify-js package

behzad888:
I have an error as 'JS_Parse_Error' on uglify-js, so update it


guybedford:
Thanks for giving this a go.

This seems to be failing some tests on `uglify.parse`. If you'd be interested in fixing these cases further that would be a great help.

guybedford:
https://github.com/systemjs/builder/pull/815 is the proposed upgrade path here for full ES support.

PR #820, opened 2017-07-15
Add uglify.output option

### LINKED ISSUE #689: builder.bundle ->  options -> uglify -> beautify should be called **output**
Since you are providing a way to pass options to Uglify directly (separate from SystemJS Builder defaults) these should be consistent with `UglifyJS.minify()`. `minify()` has no option `beautify` but has `output`^. This is passed in your code to `ast.print_to_string(outputOptions)` which is the same as what `minify` does internally* except for this name discrepancy. See lines 106-113 in lib/output.js. 

To be consistent with uglifyJS this must be changed.

* `output` has an option `beautify` inside.
* `print_to_string()` is a wrapper for function `OutputStream()` which is what `minify` invoke a directly.

CxRes:
Fixes #689! See commit message for more info.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -595,8 +595,9 @@ function processOutputOpts(options, defaults) {
   extend(opts, options);
 
   opts.uglify = opts.uglify || {};
+  opts.uglify.output   = opts.uglify.output || {};  
   opts.uglify.compress = opts.uglify.compress || {};
-  opts.uglify.beautify = opts.uglify.beautify || {};
+  opts.uglify.beautify = opts.uglify.beautify || opts.uglify.output;
 
   // NB deprecated these for uglify directly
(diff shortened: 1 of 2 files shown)


guybedford:
I guess the version upgrade from here should still go through. Would it be possible to rebase that?

CxRes:
I am not sure what to do here! The 0.15 branch is pointing to the point of branching and not the tip of the tree ie 0.15.36 tag. On my copy of the repo [CxRes/builder](https://github.com/CxRes/builder), the fix-uglify-option branch sits correctly on top of 0.15.36 tag but I don't get what github does when I place a PR from that branch that leads to this conflict????

guybedford:
This is definitely a PR against the master branch not the 0.15 branch, and the code is a diff against that fine. If you're just looking to resolve the conflicts, a `git pull --rebase` with possible conflict fixing may just be what is needed here?

PR #834, opened 2017-09-26
Fix incorrect source path normalization, when the path is either rooted in a drive letter or a file url.

thomas-darling:
This should resolve https://github.com/jspm/jspm-cli/issues/2361


guybedford:
Thanks for following up on this!

Issue #852, opened 2018-07-16
Calling builder.trace with 'externals' in options throws error.


shivijais:
`builder.trace('first.js', {
    externals: ['strings']
})`

Gives error

```
Unhandled rejection TypeError: Cannot read property 'normalize' of undefined
    at /Users/shiwani.jaiswal/thoughtspot/blink/node_modules/systemjs-builder/lib/builder.js:632:26
    at Array.map (native)
    at setExternals (/Users/shiwani.jaiswal/thoughtspot/blink/node_modules/systemjs-builder/lib/builder.js:630:32)
    at /Users/shiwani.jaiswal/thoughtspot/blink/node_modules/systemjs-builder/lib/builder.js:428:12
```

This is happening due to wrong context being passed to setExternals method.

PR #853, opened 2018-07-24
Pass correct context to setExternal method while building a trace

shivijais:
Fixes: #852


guybedford:
Nice, thanks for posting a fix.

Issue #726, opened 2016-10-25
Use Babili for minification


guybedford:
We can add a step to the compiler at https://github.com/systemjs/builder/blob/master/compilers/compiler.js, that returns `source, sourceMap, minifiedSource, minifiedSourceMap` by running another compile phase on the same AST after the initial unminified compile. We can then cache all four along with the load record and solve the problem of having granular caching of minification as well (https://github.com/systemjs/builder/issues/209).

If anyone is interested in working on this contributions would be really really amazing.

timfish:
I had a quick look at what would happen if I replace `minify()` in `output.js` with some rather hacky code calling the latest `uglify-es`:
```javascript
function minify(output, fileName, mangle, uglifyOpts) {
  var uglify = require('uglify-es');

  let files = {};
  files[fileName] = output.source;

  var result = uglify.minify(files, {
    sourceMap: {
      content: output.sourceMap,
      url: fileName + ".map"
    }
  });

  output.source = result.code;
  output.sourceMap = result.map;

  return output;
}
```

All the tests pass apart from these two:
```shell
  106 passing (16s)
  1 pending
  2 failing

  1) Test tree builds - Babel SFX tree build:
     Error: Phantom test failed test/test-sfx.html failed.
      at Error (native)
      at ChildProcess.<anonymous> (C:\Users\tim\Documents\Repos\systemjs-builder\test\test-build.js:28:16)
      at maybeClose (internal/child_process.js:877:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)

  2) Test tree builds - TypeScript SFX tree build:
     Error: Phantom test failed test/test-sfx.html failed.
      at Error (native)
      at ChildProcess.<anonymous> (C:\Users\tim\Documents\Repos\systemjs-builder\test\test-build.js:28:16)
      at maybeClose (internal/child_process.js:877:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

I tried these changes on v0.15.x as that's the version which `aurelia-bundler` is still targeting but I couldn't actually get the bundle to load in my app. The new bundle is calling` System.registry.get` in places where previously it called `System.get` and falls over quickly. I'm guessing I haven't worked out which SystemJS version goes with which bundler version. For various reasons I'm still stuck on systemjs@0.19.

guybedford:
@timfish amazing, thanks for giving this a go! It's been sitting here way to long just waiting to be picked up by someone...

[...]

PR #857, opened 2018-08-23
Use terser (uglify-js fork) to support minification of more permissive babel presets

fdintino:
Currently, only ES5 javascript can be minified, due to the reliance on uglify-js. Terser is the actively maintained fork of what used to be uglify-es, the harmony branch for uglify-js@3. By using terse the systemjs builder is able to support a broader range of ECMAScript syntax.

There were significant API changes between uglify-js@2 and uglify-js@3, and so also with terser, but the updates required in the systemjs builder minify function were fairly straightforward. This PR hews more closely to the original code than #815 does, and consequently I think it avoids regressing any features (for instance, the ability to include sourceContents in source maps or to control the comment stripping is retained).

As browsers continue to support more features beyond ES5, and because babel recommends use of the 'env' preset, it will become increasingly likely that users' babel transpilation settings will cause errors when trying to minify with jspm and friends.

As part of this PR I've added a test suite that fails before this fix and passes afterwards.

refs #815, #726

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -4,9 +4,21 @@ var fs = require('fs');
 var Promise = require('bluebird');
 var asp = require('bluebird').promisify;
 var extend = require('./utils').extend;
+var terser = require('terser');
 
 var fromFileURL = require('./utils').fromFileURL;
 var toFileURL = require('./utils').toFileURL;
+
+// Ugly hack to get around terser's strange code loading and export system
+// (inherited from uglify-js), and the fact that their SourceMap wrapper is
+// not exported
+var TerserSourceMap = new Function('MOZ_SourceMap', "exports", "require", function() {
+  var code = ['terser/lib/utils.js', 'terser/lib/sourcemap.js'].map(function(file) {
+    return fs.readFileSync(require.resolve(file), 'utf8');
+  });
+  code.push('return SourceMap;');
+  return code.join('\n');
+}())(require('source-map'), terser, require);
 
 function countLines(str) {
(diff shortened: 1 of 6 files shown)


asapach:
Unfortunately `uglify-es` is no longer supported: mishoo/UglifyJS2#3156
You can try this instead: https://github.com/fabiosantoscode/terser

fdintino:
@guybedford I know this project won't be receiving many (if any) updates now that jspm 2 is out. But is there any chance this might be merged in?

guybedford:
Sure, let's do it.

guybedford:
Published in 0.16.14.

@fdintino if you are relying on this and would like to be a co-maintainer of this project I'd be happy to add you. Just let me know.

guybedford:
Unfortunately it seems the internal terser import is broken in modern terser, so I've locked the terser version to 3.8.1 and released this in 0.16.15.

fdintino:
Ah, I'm sorry about that. I should have checked that it still worked with the most recent version of terser before asking for it to be merged. I'm glad that Terser's source map class is exposed publicly (so  that it shouldn't need the hacky pseudo-import that this PR uses). I can take a stab at that and open another PR to have it work with modern terser.

I'd be happy to be a co-maintainer. At some point I'll be upgrading my projects to jspm 2, but at least for the moment I'm still using jspm 0.17 and systemjs-builder.

guybedford:
@fdintino sure, I've added you as a collaborator. If you want to put together an updated PR I'll gladly review further and will aim to not let you wait quite so long :) Thanks for your work here.

